PR #7, opened 2018-07-16
拼写错误：`scr` 应该是 `src`

boliza:
拼写错误：`scr` 应该是 `src`


rootsongjc:
@boliza That's really a typo, thanks for your PR.

PR #68, opened 2018-07-20
Add crosscompile && fix  Newcluster && fix decode-error

Lingtaonju:


Lingtaonju:
Add crosscompile && fix  Newcluster && fix decode-error

PR #118, opened 2018-07-31
unit-test should not run code in test, fix fuzzy test

nejisama:


nejisama:
make sure http.Client reuse the connection, two things should do:
1. Read All Reponse , for example:  ioutil.ReadAll(resp.Body)
2. Resp.Body.Close()

Issue #147, opened 2018-08-08
worker pool might panic due to the incorrect event processing order


neverhook:
![image](https://user-images.githubusercontent.com/41352674/43831636-5c15be40-9b37-11e8-9460-ea72976c6f75.png)

neverhook:
fxied by #149

PR #149, opened 2018-08-08
remove event priority processing to avoid race condition bug; use str…

neverhook:
…eam state to help judging if stream event should be processed or not


neverhook:
fix #147

neverhook:
The root cause was the **incorrect consume order of stream events.**

In normal case, we assume that stream have 6 events during its lifecycle(3, 4 may exchange order if corresponding protocol supports that headers could be sent individually):

1. start
2. downstream header receive
3. downstream data receive
4. upstream header receive
5. upstream data receive 
6. stop

There are 2 possible cases could produce the #147 problem. Let's see.

# Job queue flush 

## Desc

In previous version, we have a buffered chan and slice-based job queue for each shard.
```go
type shard struct {
	sync.Mutex

	index        int
	respawnTimes uint32

	// job
	jobChan  chan interface{}
	jobQueue []interface{}

	// control
	ctrlChan  chan interface{}
	ctrlQueue []interface{}
}
```
We supposed that job event would be sent to the channel if possible, otherwise it would be append into the job queue, waiting for queue flush. Codes as following:
```go
		select {
		case shard.jobChan <- job:
		default:
			shard.jobQueue = append(shard.jobQueue, job)
			// schedule flush once
			if atomic.CompareAndSwapUint32(&p.schedule, 0, 1) {
				p.flush()
			}
		}
```
And the real flush staff is running in another seperate goroutine:
```go
func (p *shardWorkerPool) flush() {
	go func() {
               // ...flush logic
	}()
}
```
So consider this situation:
1. chan buffer is full
2. stream x downstream header event is appended into the job queue
3. worker goroutine takes some event from the job chan, so now chan slots is available.
4. if new job event comes before the flush staff done, for example, stream x downstream data event is offered, then is would be directly sent to the job chan, although the header event is still in job queue.

So in conclusion, the select statement workaround can not ensure the FIFO order. And our code would panic if processing data event before header event was handled.

## Fix code
just add judgement for the queue size, if it is not empty, jobs must be appended into the queue, not the chan.
```go
	if len(shard.jobQueue) == 0 && cap(shard.jobChan) > len(shard.jobChan) {
		shard.jobChan <- job
	} else{
		shard.jobQueue = append(shard.jobQueue, job)
		// schedule flush if
		if atomic.CompareAndSwapUint32(&p.schedule, 0, 1) {
			p.flush()
		}
	}
```
# No-guaranteed double select statement for priority chan read

## Desc
As we known, golang's buffered chan is  bounded and without priority support. In this case, we use code below to acheive the goal of **priority chan read**. 
```go
	for {
		var event interface{}

		// control event process first
		select {
		case event = <-ctrlChan:
			eventProcess(streamMap, event)
			// next loop
			continue
		default:
		}

		// job & ctrl event process
		select {
		case event = <-ctrlChan:
		case event = <-jobChan:
		}

		eventProcess(streamMap, event)
	}
```
This pattern is misused widely. Like the [first answer for this question](https://stackoverflow.com/questions/11117382/priority-in-go-select-statement-workaround)

This is actually not an entirely correct solution, since it is possible for both queues to receive data in a single context switch. The behavior of select when multiple queues are ready is indeterminate (pseudo-random).

We got the panic  under this case:
1. Control chan is clear. so we hit the default branch of the first select.
2. Another goroutine offer two events: stream x start event(control chan) and downstream header receive event(job chan).
3. Now our code runs to the second select statment.But which event would be processd first is pseudo-random. If header event processed before start event, panic occurs.

## Fix code

The originate of two priority chan design is that we hopes some events, which might affect the process flow(like reset event), could be processed before other job event. For example, header event processing may produce a reset event.If we post the reset event into the same chan buffer with data event, then data event would be processed before reset event. This is dangerous because some resource is unavailable since reset event is emitted. So we wanna use two different chan to clarify the event priority.

But this pattern is incorrect. So we use state flag(downstream.downstreamReset) to indicate the event process logic whether incoming stream event should be processed or just ignored.
```go
if done, ok := streamMap[e.streamID]; ok && !(done || streamProcessDone(e.stream)) {
	switch e.direction {
	case Downstream:
		e.stream.ReceiveHeaders(e.headers, e.endStream)
	case Upstream:
		e.stream.upstreamRequest.ReceiveHeaders(e.headers, e.endStream)
	default:
		e.stream.logger.Errorf("Unknown receiveHeadersEvent direction %s", e.direction)
	}
	streamMap[e.streamID] = streamMap[e.streamID] || streamProcessDone(e.stream)
}

func streamProcessDone(s *downStream) bool {
	return s.upstreamProcessDone || s.downstreamReset == 1
}
```

PR #154, opened 2018-08-12
fix integrate test needs init stream

nejisama:
ci check have bugs, sometimes it will not trigger integrate test check.
after merge #152 , it is needs to import stream/${protocol} to init


Lingtaonju:
maybe mosn.go also need import these packages to init

PR #175, opened 2018-08-25
Remove HSF from feature list in README.

khotyn:
Remove HSF from feature list in README, HSF is not an open source protocol, should not listed or supported in open source version of SOFAMosn.


khotyn:
Related issue #177

Issue #181, opened 2018-08-27
ActiveRequests has data race when closed


dawxy:
Data race causes panic when using the original post request without the HTTP header x-mosn-method.
```
panic: interface conversion: interface {} is nil, not *http.activeRequest

goroutine 4305471 [running]:
github.com/alipay/sofa-mosn/pkg/stream/http.(*codecClient).OnEvent(0xc42040c090, 0xe11b49, 0xb)
	/mnt/d/GoPath/src/github.com/alipay/sofa-mosn/pkg/stream/http/codecclient.go:154 +0x2ff
github.com/alipay/sofa-mosn/pkg/stream/http.(*clientStream).doSend(0xc42720cf60)
	/mnt/d/GoPath/src/github.com/alipay/sofa-mosn/pkg/stream/http/stream.go:329 +0xf6
created by github.com/alipay/sofa-mosn/pkg/stream/http.(*clientStream).endStream
	/mnt/d/GoPath/src/github.com/alipay/sofa-mosn/pkg/stream/http/stream.go:303 +0x3f
```

PR #182, opened 2018-08-27
fix: #181 Data race

dawxy:
fix:#181


ldxdl:
Because the code you changed is pretty important, pls provide more details about your the correctness of your code change, including :
+ unit test result
+ integration test result
+ fuzzy test result
+ goimports/golint results

dawxy:
@ldxdl  
Can't we use "go test -race" in unit tests? 
I see some other DATA RACE warnings.

ldxdl:
Good point, @nejisama pls pay attention on race detection in test phase, and check and fix it in 0.2.0 testing.

dawxy:
Then I cancel the -race parameter first, then only add the unit test, because I don't know how to add the integration test and fuzzy test.

PR #190, opened 2018-08-27
Fix #183

### LINKED ISSUE #183: Invalid filed content fill on parseFilterChains
Config instance fill to v2.FilterConfig.Config, but v2.Config expected

kevin21th:
add parameter isV2 to NetworkFilterChainFactory in order to distinguish the format of the map[string]interface{}


nejisama:
needs to add local test for xds

PR #212, opened 2018-08-30
fix socket fd blocking

### LINKED ISSUE #210: Origin socket fd got blocked in non-netpoll mode
We store the raw connection fd copy to  support netpoll mode for tls conn.These codes work no matter what mode you choose.
```go
	var file *os.File
	// store fd for further usage
	if tc, ok := rawc.(*net.TCPConn); ok {
		file, err = tc.File()
		if err != nil {
			return err
		}
	}
```
The File() method would cause the fd in blocking mode:
```
func (c *conn) File() (f *os.File, err error) {
	f, err = c.fd.dup()
	if err != nil {
		err = &OpError{Op: "file", Net: c.fd.net, Source: c.fd.laddr, Addr: c.fd.raddr, Err: err}
	}
	return
}

func (fd *netFD) dup() (f *os.File, err error) {
	ns, err := dupCloseOnExec(fd.pfd.Sysfd)
	if err != nil {
		return nil, err
	}

	// We want blocking mode for the new fd, hence the double negative.
	// This also puts the old fd into blocking mode, meaning that
	// I/O will block the thread instead of letting us use the epoll server.
	// Everything will still work, just with more threads.
	if err = syscall.SetNonblock(ns, false); err != nil {
		return nil, os.NewSyscallError("setnonblock", err)
	}

	return os.NewFile(uintptr(ns), fd.name()), nil
}
```
And in netpoll-mode, setNonblock was invoked so the socket fd become non-blocking again.
```go
// Handle creates new Desc with given file and event.
// Returned descriptor could be used as argument to Start(), Resume() and
// Stop() methods of some Poller implementation.
func HandleFile(file *os.File, event Event) (*Desc, error) {
	desc := NewDesc(file.Fd(), event)

	// Set the file back to non blocking mode since conn.File() sets underlying
	// os.File to blocking mode. This is useful to get conn.Set{Read}Deadline
	// methods still working on source Conn.
	//
	// See https://golang.org/pkg/net/#TCPConn.File
	// See /usr/local/go/src/net/net.go: conn.File()
	if err := setNonblock(desc.fd(), true); err != nil {
		return nil, os.NewSyscallError("setnonblock", err)
	}

	return desc, nil
}
```
But this doesn't apply to the non-nepoll-mode, so it still remains in blocking mode, which cause an huge impact on performance.

neverhook:

FIX #210 

### Solutions
Only get socket fd when needed(conenction IO enable && UseNetpollMode), and it must be ahead of tls conn wrapper works. So we adjust the tlsMng from listener to activeListener

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -35,6 +35,7 @@ import (
 	"github.com/alipay/sofa-mosn/pkg/log"
 	"github.com/alipay/sofa-mosn/pkg/network"
 	"github.com/alipay/sofa-mosn/pkg/types"
+	"github.com/alipay/sofa-mosn/pkg/tls"
 )
 
 // ConnectionHandler
@@ -191,7 +192,7 @@ func (ch *connHandler) AddOrUpdateListener(lc *v2.ListenerConfig, networkFilters
 
 		l := network.NewListener(lc, logger)
 
-		al = newActiveListener(l, logger, als, networkFiltersFactories, streamFiltersFactories, ch, listenerStopChan, lc.DisableConnIo)
+		al = newActiveListener(l, lc, logger, als, networkFiltersFactories, streamFiltersFactories, ch, listenerStopChan)
 		l.SetListenerCallbacks(al)
 		ch.listeners = append(ch.listeners, al)
 	}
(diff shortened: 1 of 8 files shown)


pkg/server/handler.go
```diff
@@ -370,12 +372,33 @@ func newActiveListener(listener types.Listener, logger log.Logger, accessLoggers
 	al.statsNamespace = types.ListenerStatsPrefix + strconv.Itoa(listenPort)
 	al.stats = newListenerStats(al.statsNamespace)
 
+	mgr, err := tls.NewTLSServerContextManager(lc, listener, logger)
+	if err != nil {
+		logger.Fatalf("create tls context manager failed, %v", err)
```
ldxdl: Do not use fatal log here, newActiveListener may be called on dynamic resource case

ldxdl:
Copied fd should be close on connection close.

PR #231, opened 2018-09-04
fix #227 use uint32 as source

### LINKED ISSUE #227: streamID use int32 does not match rpc request id uint32
request id range is uint32, but stream id use int32, may be out of range and trigger panic in proxy/event.go

nejisama:

#227 

### Sign the CLA
Make sure you has singed the [CLA](https://www.clahub.com/agreements/alipay/sofa-mosn)

### Solutions

request id in sofa rpc have 4 bytes, should use uint32 to represent

### UT result

UT test pass, no new unit test

### Benchmark

No Benchmark

### Code Style

+ Show `Golint` result

### DIFF
--- a/pkg/proxy/event.go
+++ b/pkg/proxy/event.go
@@ -36,12 +36,12 @@ type streamEvent struct {
 	stream    *downStream
 }
 
-func (s *streamEvent) Source() int {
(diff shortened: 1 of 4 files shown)


neverhook:
LGTM

PR #248, opened 2018-09-10
fix: errorf format error

dawxy:
Fix errorf format error.


nejisama:
go vet can only be used as a reference, and can not be forced to check

dawxy:
fixed

dawxy:
@nejisama 
Conflict has been resolved.

PR #360, opened 2018-11-08
feature: allow append duplicate listener in the same connection

### LINKED ISSUE #359: bug: AddConnectionEventListener allow duplicate listener
### Describe the bug

The method  `AddConnectionEventListener` 、`AddBytesReadListener` and `AddBytesReadListener`  as the following, allow duplicate listener. Call `AddConnectionEventListener` twice with same arguments and we will see the `conn.connCallbacks` grows

https://github.com/alipay/sofa-mosn/blob/9f06247cb6b37a03631a86fc05c7fda6610404c5/pkg/network/connection.go#L683-L725

should we  allow duplicate listener?, see the reason as the below:

In go, we can compare the interface if they have identical dynamic types and equal dynamic values or if both have value nil.

But the AddBytesSentListener and AddBytesReadListener tend to append function rather than interface and function values are not comparable.

So it's hard to guarantee above 3 method's semantics (disallow duplicate listener). It's okay to allow duplicate listeners to append. The caller should know what they do.

### Expected behavior
the length of `conn.connCallbacks` should be 1

### Actual behavior
the length of `conn.connCallbacks` was 2

### Steps to reproduce
Call `AddConnectionEventListener` twice with same arguments

### Environment
9f06247cb6b37a03631a86fc05c7fda6610404c5

detailyang:

Try to fix #359

### Sign the CLA
Yes

### Solutions

The Interface is comparable and if they have identical dynamic types and equal dynamic values or if both have value nil. But the function value cannot compare. So It's hard to guarantee the same behavior between `AddConnectionEventListener` 、 `TestAddBytesReadListener` and `TestAddBytesSendListener`

IMHO, Allowing duplicate listener in the same connection is okay and the caller should know what they do.

## UT result
````bash
❯ go test -v ./pkg/network/...                                                                                                                         
=== RUN   TestAddConnectionEventListener
--- PASS: TestAddConnectionEventListener (0.00s)
=== RUN   TestAddBytesReadListener
--- PASS: TestAddBytesReadListener (0.00s)
=== RUN   TestAddBytesSendListener
--- PASS: TestAddBytesSendListener (0.00s)
PASS
ok  	github.com/alipay/sofa-mosn/pkg/network	0.022s
````

### Benchmark
N/A

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/network/connection_test.go
+++ b/pkg/network/connection_test.go
@@ -1,7 +1,6 @@
 package network
 
 import (
-	"math/rand"
 	"testing"
 
 	"github.com/alipay/sofa-mosn/pkg/log"
@@ -11,10 +10,6 @@ import (
 type MyEventListener struct{}
 
 func (el *MyEventListener) OnEvent(event types.ConnectionEvent) {}
-
-func randN() int {
-	return rand.Intn(1024) + 1
-}
 
 func TestAddConnectionEventListener(t *testing.T) {
 	logger, err := log.NewLogger("stdout", log.INFO)
@@ -26,7 +21,7 @@ func TestAddConnectionEventListener(t *testing.T) {
 		logger: logger,
 	}
 
-	n := randN()
+	n := 1024
 	for i := 0; i < n; i++ {
 		el0 := &MyEventListener{}
 		c.AddConnectionEventListener(el0)
@@ -47,7 +42,7 @@ func TestAddBytesReadListener(t *testing.T) {
 		logger: logger,
 	}
 
-	n := randN()
+	n := 1024
 	for i := 0; i < n; i++ {
 		fn1 := func(bytesRead uint64) {}
 		c.AddBytesReadListener(fn1)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/network/connection_test.go [hunk 1]
```diff
@@ -0,0 +1,80 @@
+package network
```
nejisama: add license

pkg/network/connection_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/alipay/sofa-mosn/pkg/log"
+	"github.com/alipay/sofa-mosn/pkg/types"
+)
+
+type MyEventListener struct{}
+
+func (el *MyEventListener) OnEvent(event types.ConnectionEvent) {}
+
+func randN() int {
+	return rand.Intn(1024) + 1
```
nejisama: global rand needs to set rand.Seed(time.Now().UnixNano())， or it will be use Seed as 0. 
At this test, can use a const N, no needs to use rand

nejisama:
@detailyang  pls fix it, and we will merge it

pkg/network/connection_test.go
detailyang: sorry, it should be inside fuzzy tests

PR #380, opened 2018-11-14
bugfix: avoid memory leak in downstream (http1 or http2)

### LINKED ISSUE #379: memory leak when dowmstream is  http1 or http2
### Describe the bug

memory leak when the protocol of downstream is HTTP1 or HTTP2. Since the io loop is hijacked by the fasthttp or golang's http2 and we have no chance to bind the close|reset event to OnEvent callback which will free the activeConnection and reduce the num connections. See the code below:

https://github.com/alipay/sofa-mosn/blob/e2fafeabed8f04ecd8ff6a0d7191a7bb6531cf5e/pkg/server/handler.go#L438-L468

when returning from the `nfcf.CreateFilterChain(ctx, al.handler.clusterManager, filterManager)` and the protocol is http1 or http2. It means the connection was finalized (close or reset) and no need to insert it to the listeners' list which will no chance to free.

### Expected behavior
Do not leak

### Actual behavior
Leak

### Steps to reproduce
1.  copy the config like the following

https://gist.github.com/detailyang/4480861230f753b02df4cf6f309bcb99

2. run the backend server in 8080 whatever (I use node), `node index.js 8080`

````node
const http = require('http');
const port = process.argv.length > 2 ? process.argv[2] : 8080;

const handler = (request, response) => {
  response.end('Hello World!\r\n')
}

const server = http.createServer(handler)

server.keepAliveTimeout = 30000; // enable keepalive for test
server.listen(port, (err) => {
  if (err) {
    return console.log('something bad happened', err)
  }
  console.log(`server is listening on ${port}`)
})
````

3. send thousands of  requests to mosn whatever ( I use wrk), `wrk http://127.0.0.1:2046 -c 1 -t 1 -d 30s -H "Connection: Close"` or `wrk http://127.0.0.1:2047 -c 1 -t 1 -d 30s -H "Connection: Close"` (fake http2) 

4. wait for a minute and see the RSS of mosn process.

### Environment
e2fafea

detailyang:

Try to fix #379 

### Sign the CLA
Yes

### Solutions
Since the fasthttp and http2 will hijack the io loop. We have no chance to bind the close|reset event to our event callback. We could think when we yield from the fasthttp and http2 io loop, the connection is finalized and we can return directly and reduce the number of connections.

### UT result
N/A

### Benchmark
N/A

### Code Style
N/A

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -461,7 +461,7 @@ func (al *activeListener) OnNewConnection(ctx context.Context, conn types.Connec
 	// and we can return directly.
 	if al.disableConnIo {
 		atomic.AddInt64(&al.handler.numConnections, -1)
-		al.logger.Debugf("new downstream connection %d closed", conn.ID())
+		al.logger.Debugf("downstream connection %d was closed", conn.ID())
 		return
 	}
 
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -436,6 +436,10 @@ func (al *activeListener) OnAccept(rawc net.Conn, handOffRestoredDestinationConn
 }
 
 func (al *activeListener) OnNewConnection(ctx context.Context, conn types.Connection) {
+	atomic.AddInt64(&al.handler.numConnections, 1)
+
+	al.logger.Debugf("new downstream connection %d accepted", conn.ID())
+
 	//Register Proxy's Filter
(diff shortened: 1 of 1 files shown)


neverhook:
HTTP/1 and HTTP/2 support IO management in recent changes: #393  #396. Consider out-dated.

Issue #232, opened 2018-09-05
support convert FaultInject streamfilter from xds


kevin21th:
convert faultinject config from xds to mosn api/v2

nejisama:
fix #384

Issue #370, opened 2018-11-13
only one of the stream filters can be run


nejisama:
code in proxy/streamfilters bug:
```
func (s *downStream) runAppendHeaderFilters(filter *activeStreamSenderFilter, headers types.HeaderMap, endStream bool) bool {
        var index int 
        var f *activeStreamSenderFilter
        if filter != nil {
                index = filter.index + 1 
        }   
        for ; index < len(s.senderFilters); index++ {
                f = s.senderFilters[index]
                s.filterStage |= EncodeHeaders
                status := f.filter.AppendHeaders(headers, endStream)
                s.filterStage &= ^EncodeHeaders
                if status == types.StreamHeadersFilterStop {
                        f.stopped = true
                        return true
                }   
                f.headersContinued = true
                return false
        }   
        return false
}
```
no matter what result returned from f.filter.AppendHeaders,  the function will return, the for loop just run once

PR #384, opened 2018-11-15
Feature 232

nejisama:

#232 

### Sign the CLA

Done

#### feature

+ Support FaultInject as stream filter, can inject status and delay with percentage
+ Support inject conditions: the matched upstream or headers

#### JSON Config

```
{
    "delay":{
        "fixed_delay":"1s",
        "percentage": 100
    }
    "abort": {
        "status": 500,
        "percentage": 100
    },
    "upstream_cluster": "clustername",
    "headers": [
        {"name":"service","value":"test","regex":false},
        {"name":"user","value":"bob", "regex":false}
    ]
}
```
#### Changes 

+ StreamReceiverFilterCallbacks add function:  SendHijackReply.  downstream export sendHijackReply for it
+ package router export function getRouterHeaders as GetRouterHeaders; filter use router.ConfigUtilityInst.MatchHeaders to check wether the header is matched for fault inject
+ downstream run ReceiveStreamFilters after route is choosed

### UT result
pass

### Benchmark

no

### Code Style
+ Show `Golint` result


nejisama:
#370  block integrate with istio use fault inject

PR #386, opened 2018-11-16
optimize handler chain

nejisama:
Optimize handler chain

1.  RouterHandler returns Status instead of bool, can stop directly (route not found) 
2. HandlerChain keeps the clustersnapshot, returns to proxy,  so the Handler's snapshot is same as proxy used

### DIFF
--- a/pkg/router/handlerchain.go
+++ b/pkg/router/handlerchain.go
@@ -29,26 +29,39 @@ func init() {
 
 // RouteHandlerChain returns first available handler's router
 type RouteHandlerChain struct {
-	ctx      context.Context
-	handlers []types.RouteHandler
-	index    int
+	ctx            context.Context
+	handlers       []types.RouteHandler
+	clusterManager types.ClusterManager
+	index          int
 }
 
-func NewRouteHandlerChain(ctx context.Context, handlers []types.RouteHandler) *RouteHandlerChain {
+func NewRouteHandlerChain(ctx context.Context, clusterManager types.ClusterManager, handlers []types.RouteHandler) *RouteHandlerChain {
 	return &RouteHandlerChain{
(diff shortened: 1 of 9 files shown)


pkg/router/handlerchain.go
```diff
@@ -29,26 +29,39 @@ func init() {
...
+	switch status {
+	case types.HandlerAvailable:
+		return snapshot, handler.Route()
+	case types.HandlerNotAvailable:
+		hc.clusterManager.PutClusterSnapshot(snapshot)
+		return hc.DoNextHandler()
+	case types.HandlerStop:
+		return nil, nil
+	default:
+		panic("unexpected handler status")
```
ldxdl: 这里panic需要再仔细考虑一下，如果没有router，是否重新做路由选择或对downstream返回404之类

pkg/types/route.go
```diff
@@ -56,9 +56,18 @@ type RouterManager interface {
...
+const (
+	HandlerAvailable HandlerStatus = iota
+	HandlerNotAvailable
+	HandlerStop
+)
+
 // RouteHandler is an external check handler for a route
 type RouteHandler interface {
-	IsAvailable(context.Context) bool
+	IsAvailable(context.Context, ClusterSnapshot) HandlerStatus
```
ldxdl: 缺少注释

pkg/router/handlerchain.go
```diff
@@ -29,26 +29,39 @@ func init() {
...
 	}
 	return hc.DoNextHandler()
```
ldxdl: 这里感觉是dead code~ 建议用go vet之类的先检测一把

ldxdl:
LGTM

PR #415, opened 2018-12-07
groups network pkg imports and fix docs mistakes

sevennt:


sevennt:
I have already signed CLA.

PR #481, opened 2019-01-03
support GOMAXPROCS env variable

taoyuanyuan:
### Solutions
GOMAXPROCS 优先级最高，其次配置项"processor"， 最后是默认值 runtime.NumCPU()

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


ldxdl:
LGTM

PR #620, opened 2019-04-10
Enhance log with traceid

neverhook:
### Solutions
- provide default log format with **connection**, **trace id** for proxy scenario.
- trace span would be built at stream layer, which is earlier than before(proxy layer)


pkg/log/proxylog.go
```diff
@@ -0,0 +1,127 @@
...
+	if l.Logger.disable {
+		return
+	}
+	if l.level >= ERROR {
+		s := l.formatter(ctx, ErrorPre, format)
+		l.Logger.Printf(s, args...)
+	}
+}
+
+func (l *proxyLogger) Fatalf(ctx context.Context, format string, args ...interface{}) {
```
nejisama: Fatal的实现似乎有问题

neverhook: fixed

PR #648, opened 2019-04-24
fix: resource default matcher headers relation misuse

yittg:

It's a simple fix about misuse of `HeadersRelation`/`ParamsRelation` in resource defaultmatcher.

### Solutions
use `HeadersRelation` as relation for `matchHeaders` instead of `ParamsRelation`.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


yittg:
@nejisama would you take a look about this PR, :)

PR #716, opened 2019-06-24
fix(http): deal with stream reset/destroy properly

yittg:
just close http client upstream connection when downstream reset,
or the connection should wait the last response if exist even not pooled immediately,
and the fix in PR: #698 will lead to connections leak.

connections closed locally should not spread read/write error
which will lead to another close.

### DIFF
--- a/pkg/stream/http/stream.go
+++ b/pkg/stream/http/stream.go
@@ -116,6 +116,7 @@ type streamConnection struct {
 
 	conn              types.Connection
 	connEventListener types.ConnectionEventListener
+	resetReason       types.StreamResetReason
 
 	bufChan    chan types.IoBuffer
 	connClosed chan bool
@@ -218,7 +219,11 @@ func (conn *clientStreamConnection) serve() {
 		if err != nil {
 			if s != nil {
 				log.Proxy.Errorf(s.connection.context, "[stream] [http] client stream connection wait response error: %s", err)
-				s.ResetStream(types.StreamRemoteReset)
+				reason := conn.resetReason
+				if reason == "" {
+					reason = types.StreamRemoteReset
+				}
+				s.ResetStream(reason)
 			}
 			return
 		}
@@ -280,6 +285,7 @@ func (conn *clientStreamConnection) ActiveStreamsNum() int {
 func (conn *clientStreamConnection) Reset(reason types.StreamResetReason) {
 	close(conn.bufChan)
 	close(conn.connClosed)
+	conn.resetReason = reason
 }
 
 // types.ServerStreamConnection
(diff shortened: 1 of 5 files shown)


yittg:
@neverhook would you have a look, thanks.

pkg/stream/http/stream.go
```diff
@@ -280,6 +280,9 @@ func (conn *clientStreamConnection) ActiveStreamsNum() int {
 func (conn *clientStreamConnection) Reset(reason types.StreamResetReason) {
 	close(conn.bufChan)
 	close(conn.connClosed)
+	if conn.stream != nil {
+		conn.stream.ResetStream(reason)
+	}
 }
```
neverhook: Can you explain these 3 lines? I can't get the point.

yittg: stream on connection should respond connection reset event
mosn http1 client treat all err from response read as remote reset before, cause it can not distinguish io error reading from response
actually all io error will callback listener with eventType/reason(maybe not much clear now), here is on the callback path.

neverhook: I'm worry about the concurrency problem.
Mosn http1 impls an extra goroutine, called 'serve', to handle the codec and request handling. The 'serve' goroutine may be dealing with the upstream response data while downstream connection closing. Invoke the stream.Reset from downstream IO goroutine(or other goroutine which invoke the connection.Close) directly can change the stream's state or data, which may be used by the 'serve' goroutine at the same time.

neverhook: So we deal the reset event by passing it through the channel, and the 'serve ' goroutine should know the event by the channel event or response read error.
----
'it can not distinguish io error reading from response' 
I think we can handle this by adding extra flag of stream. like this

```go
func (conn *clientStreamConnection) Reset(reason types.StreamResetReason) {
	close(conn.bufChan)
	close(conn.connClosed)
	atomic.Store(&conn.localReset, 1)
}

// 1. blocking read using fasthttp.Response.Read
		err := s.response.Read(conn.br)
		if err != nil {
			if s != nil {
				log.Proxy.Errorf(s.connection.context, "[stream] [http] client stream connection wait response error: %s", err)
                                 if atomic.Load(&conn.localReset) == 1 {
                                     s.ResetStream(types.StreamLocalReset)
                                 } else {
                                     s.ResetStream(types.StreamRemoteReset)
                                 }
			}
			return
		}
``
IMO, all events should be passed to the 'serve' goroutine(maybe through channel or flags).

yittg: agree with this, another commit on the way, :)

PR #721, opened 2019-07-03
Close idle

nejisama:

### Sign the CLA
Make sure you have signed the [CLA](https://www.clahub.com/agreements/alipay/sofa-mosn)

### Solutions

+ server close connection when read timeout reached the max count
+ client close (sofarpc) connection if the connection only have heartbeat on it. (do not apply yet)

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/stream/sofarpc/idlefree.go
```diff
@@ -0,0 +1,63 @@
...
+package sofarpc
+
+import (
+	"sync/atomic"
+
+	"sofastack.io/sofa-mosn/pkg/log"
+)
+
+// DefaultConnReadTimeout is 15s, default idle free is 600s
+var defaultMaxIdleCount uint32 = 40
```
taoyuanyuan: Configurable

PR #862, opened 2019-11-11
fix default pid file path as conf

nejisama:
change pid file from mosn/logs to mosn/conf


taoyuanyuan:
modify pid file use config

PR #885, opened 2019-11-25
pkg/xds/v2: add StreamClientMutex.

chainhelen:
`=` assignment is not atomic in golang and partial assignments that may
occur probabilistically. `ADSClient.StreamClient` are both read/written
in two goroutinues, so adding `mutex` to guarantee concurrent security.

No.

### Sign the CLA
Done. 

### Solutions
`=` assignment is not atomic in golang and partial assignments that may
occur probabilistically. `ADSClient.StreamClient` are both read/written
in two goroutinues, so adding `mutex` to guarantee concurrent security.

### UT result
Probabilistically and theoretically.

### Benchmark
Not involve.

### Code Style
Done in goland.
+ Show `Golint` result
Print nothing.


chainhelen:
Travis ci failed, but it seems that doesn't involve my modification.

PR #929, opened 2020-01-06
fix request exception caused by no content-length header.

wangfakang:

fix request exception caused by no content-length header, issue  https://github.com/mosn/mosn/issues/926 .

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/stream/client.go
```diff
@@ -147,6 +148,15 @@ func (c *client) OnEvent(event types.ConnectionEvent) {
 
 		if c.ConnectedFlag {
 			reason = types.StreamConnectionTermination
+
+			if c.Protocol == protocol.HTTP1 {
```
taoyuanyuan: http is not visible to the stream layer

wangfakang: Yeah, In order to be compatible with exception scenarios in the HTTP  and not affect the processing logical of other protocols. so hardcode here.   
TODO
compatible with all protocols about `event` and `reason`.

PR #942, opened 2020-01-19
transferReadSendData is not check tls conn  is null

### LINKED ISSUE #941: transferReadSendData is not check tls conn  is null
### Describe the bug

func transferReadSendData(uc *net.UnixConn, c *mtls.TLSConn, buf types.IoBuffer, logger log.ErrorLogger) error {
	// send header
	s1 := buf.Len()
	s2 := c.GetTLSInfo(buf)
	err := transferSendHead(uc, uint32(s1), uint32(s2))
	if err != nil {
		return err
	}
	logger.Infof("TransferRead dataBuf = %d, tlsBuf = %d", s1, s2)
	// send read/write buffer
	return transferSendIoBuffer(uc, buf)
}

if  mtls.TLSConn is conn, process will pannic

### Environment

- MOSN Version

NameHaibinZhang:
Fixes #941 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/network/transfer.go
+++ b/pkg/network/transfer.go
@@ -411,7 +411,12 @@ func transferRecvType(uc *net.UnixConn) (net.Conn, error) {
 func transferReadSendData(uc *net.UnixConn, c *mtls.TLSConn, buf types.IoBuffer, logger log.ErrorLogger) error {
 	// send header
 	s1 := buf.Len()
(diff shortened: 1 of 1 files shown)


nejisama:
this code is useless change， GetTLSInfo will return 0 if conn is nil

PR #944, opened 2020-01-19
Feature/transferbugfix

### LINKED ISSUE #943: transferRecvHead  data length and tls length get wrong
### Describe the bug

func transferRecvHead(uc *net.UnixConn) (int, int, error) {
	buf, err := transferRecvMsg(uc, 8)
	if err != nil {
		return 0, 0, fmt.Errorf("ReadMsgUnix error: %v", err)
	}
	size := int(binary.BigEndian.Uint32(buf[0:]))
	id := int(binary.BigEndian.Uint32(buf[4:]))
	return size, id, nil
}

/**
 *  transfer read protocol
 *  header (8 bytes) + (readBuffer data) + TLS
 *
 * 0                       4                       8
 * +-----+-----+-----+-----+-----+-----+-----+-----+
 * |      data length      |     TLS length        |
 * +-----+-----+-----+-----+-----+-----+-----+-----+
 * |                     data                      |
 * +-----+-----+-----+-----+-----+-----+-----+-----+
 * |                     TLS                       |
 * +-----+-----+-----+-----+-----+-----+-----+-----+

size := int(binary.BigEndian.Uint32(buf[0:])) will get data length and tls length， so get data lenght
is wrong

### Environment

- MOSN Version

NameHaibinZhang:
Fixes #943 

get data lenght is wrong, and change 
size := int(binary.BigEndian.Uint32(buf[0:]))

to 

size := int(binary.BigEndian.Uint32(buf[:4]))

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/network/transfer.go
+++ b/pkg/network/transfer.go
@@ -411,7 +411,12 @@ func transferRecvType(uc *net.UnixConn) (net.Conn, error) {
 func transferReadSendData(uc *net.UnixConn, c *mtls.TLSConn, buf types.IoBuffer, logger log.ErrorLogger) error {
 	// send header
 	s1 := buf.Len()
-	s2 := c.GetTLSInfo(buf)
+	var s2 int
+	if c != nil {
+		s2 = c.GetTLSInfo(buf)
+	} else {
+		s2 = 0
+	}
 	err := transferSendHead(uc, uint32(s1), uint32(s2))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


nejisama:
the uint32 is 4 bytes, nothing is wrong.
if you have scene show this is wrong. please add test code

PR #953, opened 2020-02-03
Optimize some logical for readOnce.

wangfakang:
It can cause a mem-leak  when it requests a large amount of data.

### Solutions
A. buf multiplexing logic, which triggers recycling when the reusable space is no less than the free space.
B. buf expansion logic, adopt the method of segmentally slow start expansion (minimum according to 1k allocation, more than 1k and less than 4M increase at twice the speed, more than 4M increase at 1.5 times the speed).
C. Buf recovery logic. After the data on the buf is consumed, if the total size of the free buf exceeds 1M, it will be quickly recovered and the 128k buf will be pre-allocated for the next use.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/buffer/iobuffer.go
+++ b/pkg/buffer/iobuffer.go
@@ -28,7 +28,7 @@ import (
 	"mosn.io/mosn/pkg/types"
 )
 
-const AutoExpand = 1
+const AutoExpand = -1
 const MinRead = 1 << 9
 const MaxRead = 1 << 17
 const ResetOffMark = -1
@@ -376,11 +376,15 @@ func (b *IoBuffer) SetEOF(eof bool) {
 	b.eof = eof
 }
 
+//The expand parameter means the following:
+//A, if expand > 0, cap(newbuf) is calculated according to cap(oldbuf) and expand.
+//B, if expand == AutoExpand, cap(newbuf) is calculated only according to cap(oldbuf).
+//C, if expand == 0, only copy, buf not be expanded.
 func (b *IoBuffer) copy(expand int) {
 	var newBuf []byte
 	var bufp *[]byte
 
-	if expand > 0 {
+	if expand > 0 || expand == AutoExpand {
 		cap := cap(b.buf)
 		// when buf cap greater than MaxThreshold, start Slow Grow.
 		if cap < 2*MinRead {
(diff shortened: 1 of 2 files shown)


pkg/buffer/iobuffer.go
```diff
@@ -424,7 +383,17 @@ func (b *IoBuffer) copy(expand int) {
 	var bufp *[]byte
 
 	if expand > 0 {
```
taoyuanyuan: 这儿expand没有意义了，就可以直接干掉了，改成bool值。

pkg/buffer/iobuffer.go
```diff
@@ -71,91 +73,48 @@ func (b *IoBuffer) Read(p []byte) (n int, err error) {
...
-		if l != m {
-			loop = false
-		}
+		m, err = r.Read(b.buf[len(b.buf):cap(b.buf)])
 
-		if n > MaxRead {
-			loop = false
-		}
+		// Reset read deadline
+		conn.SetReadDeadline(zeroTime)
```
taoyuanyuan: 这个设置应该是多余的了

wangfakang: 是的，已经更新了。

pkg/buffer/iobuffer.go
wangfakang: 这个地方的expand还不能去掉，因为在外部使用该buffer进行Write的时，每次都认为全部写到buffer里面（比如http里面的[Write](https://github.com/mosn/mosn/blob/master/pkg/stream/http/stream.go#L179)这个里面的[Write](https://github.com/mosn/mosn/blob/master/pkg/buffer/iobuffer.go#L199)可能会进行grow）。

taoyuanyuan: 这个expend的int值是没有意义的，你看后面也没用依赖这个值来扩容，也是根据cap来自己计算的。
改成bool值，就表示是否需要扩容是不是更好理解。

wangfakang: 现在这个expand值是有2个含义的，当设置为AutoExpand时就根据buf的自身cap进行扩容，如果是非AutoExpand，则其expand的值是有含义的(至少扩容expand大小的buf)。非AutoExpand主要是针对每次扩容的buf必须超过expand的值，比如MOSN中的http里面的Write这个里面的Write会根据expand进行grow。

nejisama:
buffer is moved to github.mosn/mosn/pkg

Issue #960, opened 2020-02-06
Fix SDS Provider Nil Panic


pxzero:
### Describe the bug

When new  SDS Provider fail , AddOrUpdateListener will panic.

The Solution is handling nil and retry next time config update event.

### Environment

- MOSN Version

PR #961, opened 2020-02-06
fix sds provider return nil

pxzero:
For #960

### Solutions
The Solution is handling nil and retry next time config update event.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/mtls/secret_manager.go
```diff
@@ -86,6 +86,9 @@ func (mng *secretManager) getOrCreateProvider(cfg *v2.TLSConfig) *sdsProvider {
 		v.certificates[certName] = p
 		// set a certificate callback
 		client := GetSdsClient(cfg.SdsConfig.CertificateConfig.Config)
+		if client == nil {
```
nejisama: 这里可以补一个ERROR日志，然后后面两个return error 应该忽略掉

pxzero: fixed

PR #995, opened 2020-02-20
 Feature/router config

nejisama:
Refactory router config.

1.  add config field: routers in server, instead of the filter config "connection_manager"
2. keep the compatible config parse for "connection_manager", will be removed some versions later.
3. add a new feature controlled by featuregate: auto_config, default value is false.
4. if auto_config is enable,  update the mosn config, for example, by xds apis, will rewrite the mosn start config


netroby:
#1013 tracking this pr, if this pr getting merged, i will know

Issue #1017, opened 2020-03-04
go mod verify failed


champly:
go mod verify failed

### env

- git commit 0407b112
- golang v1.14
- darwin

### setup

run `go mod download` get this error

``` shell
$ go mod download
verifying github.com/apache/dubbo-go-hessian2@v1.3.0: checksum mismatch
        downloaded: h1:VLDQuaoFRI6Wh9QJwRka4/Y9a0sE3jJXCGYGWw5bjvE=
        go.sum:     h1:ZhQYDm8GHqIp6i53T4ZJHQBN11nAYAjxlwoVznfyvD8=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.
```

### go env

``` shell
$ go env         
GO111MODULE="auto"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/champly/Library/Caches/go-build"
GOENV="/Users/champly/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/champly/go"
GOPRIVATE=""
GOPROXY="https://goproxy.cn,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/champly/go/src/mosn.io/mosn/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/g1/2gs9v9b92fx5qjmsgcxz6qw40000gn/T/go-build051761424=/tmp/go-build -gno-record-gcc-switches -fno-common"
```
[...]

PR #1018, opened 2020-03-04
fix go mod dubbo-go-hessian2

champly:
#1017


nejisama:
the different go version looks like have different go.sum

nejisama:
mosn use golang 1.12.7, not go 1.14

PR #1020, opened 2020-03-04
Support dynamic resolve host.

wangfakang:
When the host in the cluster is configured with a domain, it only be resolved once when MOSN is started. This may cause a failure to dynamically perceive domain name record updates. 

```
  "cluster_manager":{     
     "clusters":[
	{
	  "Name": "testCluster",
	  "type": "SIMPLE",
	  "lb_type": "LB_RANDOM",
	  "max_request_per_conn": 1024,
	  "conn_buffer_limit_bytes":32768,
	  "hosts":[
	     {"address":"www.taobao.com:80"}
	  ]
	}
     ]
  }
```

### Solutions
The domain is resolved dynamically at runtime, as detailed below

* When the request comes in, it gets the resolution record from the cache, and it needs to check whether the time expires.
* If it doesn't expire, it will be used directly. If it expires, it will use the last result， start a goroutine to resolve the domain asynchronously, and update the result to the cache.
*  If cache miss, synchronous domain resolution and save the result to the cache.  If the domain is IP  and  the cache will not expire. If it is domain , the cache expiration time will be set.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/upstream/cluster/expire_map.go
```diff
@@ -0,0 +1,88 @@
...
+	"mosn.io/pkg/utils"
+	"sync"
+	"time"
+)
+
+const NeverExpire int64 = -1
+
+type expiredata struct {
+	data        interface{}
+	expiredTime int64
```
taoyuanyuan: time.Duration

pkg/upstream/cluster/host.go
```diff
@@ -137,17 +138,36 @@ func (sh *simpleHost) Health() bool {
...
 	if err != nil {
 		log.DefaultLogger.Errorf("[upstream] resolve addr %s failed: %v", addrstr, err)
 		return nil
 	}
-	AddrStore.Store(addrstr, addr)
+
+	if addr.String() != addrstr {
+		// TODO support config or depends on DNS TTL for expire time
+		// now set default expire time == 100 s, Means that after 100 seconds, the new request will trigger domain resolve.
+		AddrStore.Set(addrstr, addr, time.Duration(100)*time.Second)
```
taoyuanyuan: time.Duration(100)*time.Second  -> 100 * time.Second

wangfakang: done.

Issue #1011, opened 2020-03-03
mac osx 10.15 failed : mkdir /home/admin: operation not supported


netroby:
mac osx 10.15.3 , can not start mosn. due to reason
```

2020-03-03 12:26:23,629 [FATAL] [server] [init] initialize default logger failed : mkdir /home/admin: operation not supported

```

nejisama:
the log path is configured in mosn_config.json
not hardcode

netroby:
I mean the default config file shipped in the `github.com/mosn/mosn`, should not hard code a `/home/admin` directory.

Every one will failed to run mosn failed with mac osx(due to the path not exists, and can not create it , even with sudo .

nejisama:
all the config files in the configs directory is just an example. you can edit the config file as you need

netroby:
The default config should just open then work. do not need any modify. or the example may broken . hard to understand. why not provide us the right example.

nejisama:
fixed in #1021

netroby:
I see the Pr , thanks for great works

PR #1021, opened 2020-03-05
copy examples config into configs

nejisama:
fix config files in configs https://github.com/mosn/mosn/issues/1011


taoyuanyuan:
LGTM

PR #1028, opened 2020-03-08
feat: support catch panic for admin api

caorong:

support catch panic stack for admin api, log style is same to http/server.go

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

admin/server/server.go:37:1: exported function RegisterAdminHandleFunc should have comment or be unexported
admin/server/server.go:56:6: exported type Server should have comment or be unexported
admin/server/server.go:77:1: exported method Server.Start should have comment or be unexported

### DIFF
--- a/pkg/admin/server/server.go
+++ b/pkg/admin/server/server.go
@@ -20,6 +20,7 @@ package server
 import (
 	"fmt"
 	"net/http"
+	"runtime"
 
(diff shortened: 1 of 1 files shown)


nejisama:
The http.Server will catch the Hanlder's panic , do not need to catch it by mosn

PR #1060, opened 2020-03-24
Fix original dst filter.

wangfakang:

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/server/handler.go
```diff
@@ -435,15 +436,17 @@ func (al *activeListener) OnAccept(rawc net.Conn, useOriginalDst bool, oriRemote
...
-	}
-
 	// listener filter chain.
 	for _, lfcf := range al.listenerFiltersFactories {
 		arc.acceptedFilters = append(arc.acceptedFilters, lfcf)
 	}
 
+	if useOriginalDst {
+		arc.useOriginalDst = true
+		// TODO remove it when Istio deprecate UseOriginalDst.
```
nejisama: 是否还应该加上并且acceptedFilters是空？

wangfakang: 对于original_dst filter如果重复的话，其filter内部自身会有判断的。

PR #1065, opened 2020-03-25
support least active request LB and weight  least active reques LB

CodingSinger:
support least active request loadbalancer which always get the host with the least number of active calls.
 * If there is only one host, use the host directly;
 * if there are multiple host with same least active  request value , then choose one based on random.

### DIFF
--- a/pkg/upstream/cluster/cluster.go
+++ b/pkg/upstream/cluster/cluster.go
@@ -194,7 +194,7 @@ func (ci *clusterInfo) LbOriDstInfo() types.LBOriDstInfo {
 	return ci.lbOriDstInfo
 }
 
-func (ci *clusterInfo) LbConfig() v2.IsCluster_LbConfig{
+func (ci *clusterInfo) LbConfig() v2.IsCluster_LbConfig {
 	return ci.lbconfig
 }
 
--- a/pkg/upstream/cluster/cluster.go
+++ b/pkg/upstream/cluster/cluster.go
@@ -146,8 +146,7 @@ type clusterInfo struct {
 	lbOriDstInfo         types.LBOriDstInfo
 	tlsMng               types.TLSContextManager
 	connectTimeout       time.Duration
-	lbconfig	         v2.IsCluster_LbConfig
-
+	lbConfig             v2.IsCluster_LbConfig
 }
 
 func (ci *clusterInfo) Name() string {
@@ -195,7 +194,7 @@ func (ci *clusterInfo) LbOriDstInfo() types.LBOriDstInfo {
 }
 
 func (ci *clusterInfo) LbConfig() v2.IsCluster_LbConfig {
-	return ci.lbconfig
+	return ci.lbConfig
 }
 
 type clusterSnapshot struct {
--- a/pkg/upstream/cluster/cluster.go
+++ b/pkg/upstream/cluster/cluster.go
@@ -79,7 +79,7 @@ func newSimpleCluster(clusterConfig v2.Cluster) *simpleCluster {
 	cluster.snapshot.Store(&clusterSnapshot{
 		info:    info,
 		hostSet: hostSet,
-		lb:      NewLoadBalancer(info.lbType, hostSet),
+		lb:      NewLoadBalancer(info, hostSet),
 	})
 	if clusterConfig.HealthCheck.ServiceName != "" {
 		log.DefaultLogger.Infof("[upstream] [cluster] [new cluster] cluster %s have health check", clusterConfig.Name)
@@ -97,7 +97,7 @@ func (sc *simpleCluster) UpdateHosts(newHosts []types.Host) {
 	if info.lbSubsetInfo.IsEnabled() {
 		lb = NewSubsetLoadBalancer(info, hostSet)
 	} else {
-		lb = NewLoadBalancer(info.lbType, hostSet)
+		lb = NewLoadBalancer(info, hostSet)
 	}
 	sc.lbInstance = lb
 	sc.hostSet = hostSet
@@ -146,6 +146,8 @@ type clusterInfo struct {
 	lbOriDstInfo         types.LBOriDstInfo
 	tlsMng               types.TLSContextManager
 	connectTimeout       time.Duration
+	lbconfig	         v2.IsCluster_LbConfig
+
 }
 
 func (ci *clusterInfo) Name() string {
@@ -190,6 +192,10 @@ func (ci *clusterInfo) ConnectTimeout() time.Duration {
 
 func (ci *clusterInfo) LbOriDstInfo() types.LBOriDstInfo {
 	return ci.lbOriDstInfo
+}
+
+func (ci *clusterInfo) LbConfig() v2.IsCluster_LbConfig{
+	return ci.lbconfig
 }
 
 type clusterSnapshot struct {
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -44,6 +44,7 @@ func (edf *edfSchduler) Add(item WeightItem, weight float64) {
 }
 
 // Pick entry with closest deadline and push again
+// Note that you need to ensure the return result of weightFunc is not equal to 0
 func (edf *edfSchduler) NextAndPush(weightFunc func(item WeightItem) float64) interface{} {
 	edf.lock.Lock()
 	defer edf.lock.Unlock()
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -1,0 +1,71 @@
+package cluster
+
+import (
+	"container/heap"
+	"sync"
+)
+
+type edfSchduler struct {
+	lock  sync.Mutex
+	items PriorityQueue
+}
+
+func newEdfScheduler() *edfSchduler {
+	return &edfSchduler{}
+}
+
+// edfEntry is an internal wrapper for item that also stores weight and relative position in the queue.
+type edfEntry struct {
+	deadline float64
+	weight   uint32
+	item     interface{}
+}
+
+// PriorityQueue
+type PriorityQueue []*edfEntry
+
+func (pq PriorityQueue) Len() int           { return len(pq) }
+func (pq PriorityQueue) Less(i, j int) bool { return pq[i].deadline < pq[j].deadline }
+func (pq PriorityQueue) Swap(i, j int)      { pq[i], pq[j] = pq[j], pq[i] }
+
+func (pq *PriorityQueue) Push(x interface{}) {
+	*pq = append(*pq, x.(*edfEntry))
+}
+
+func (pq *PriorityQueue) Pop() interface{} {
+	old := *pq
+	*pq = old[0 : len(old)-1]
+	return old[len(old)-1]
+}
+
+
+// Current time in EDF scheduler.
+func (edf *edfSchduler) currentTime() float64 {
+	if len(edf.items) == 0 {
+		return 0.0
+	}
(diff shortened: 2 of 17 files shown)


pkg/upstream/cluster/loadbalancer.go
```diff
@@ -45,6 +46,7 @@ func init() {
 	}
 	RegisterLBType(types.RoundRobin, rrFactory.newRoundRobinLoadBalancer)
 	RegisterLBType(types.Random, newRandomLoadBalancer)
```
wangfakang: it should update [convertLbPolicy]( https://github.com/mosn/mosn/blob/master/pkg/xds/conv/convertxds.go#L985) function  for Istio as a configuration source scenario.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -129,5 +131,54 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+		rand:  rand.New(rand.NewSource(time.Now().UnixNano())),
+	}
+}
+
+func (lb *leastActiveLoadBalancer) ChooseHost(context types.LoadBalancerContext) types.Host {
+	healthHosts := lb.hosts.HealthyHosts()
+	if len(healthHosts) == 0 {
+		return nil
+	}
+	// The list of hosts having the same least active reqeuest value
```
wangfakang: if single host, we can return early.

CodingSinger: good advice.

pkg/upstream/cluster/loadbalancer_test.go
```diff
@@ -0,0 +1,35 @@
+package cluster
```
taoyuanyuan: could you add a test case of benchmark ？

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -129,5 +131,58 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+	}
+	// exactly one healthy host, return this host directly
+	if len(healthHosts) == 1 {
+		return healthHosts[0]
+	}
+	// The list of hosts having the same least active request value
+	candicate := make([]types.Host, 0, len(healthHosts))
+	// The least active request value of all hosts
+	leastActive := int64(math.MaxInt64)
+	for _, host := range healthHosts {
```
taoyuanyuan: If there are many hosts, it's slow.
We could begin host loop with a random index, and return it directly if the count is zero.

CodingSinger: return it directly if the count is zero is a good idea, But is it really necessary to loop hosts with a random index?

pkg/upstream/cluster/loadbalancer_test.go
CodingSinger: done.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -129,5 +131,58 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+		} else if active == leastActive {
+			candicate = append(candicate, host)
+		}
+	}
+	//  exactly one host, return this host directly
+	if len(candicate) == 1 {
+		return candicate[0]
+	}
+	// choose one candicate based on the random
+	return candicate[lb.rand.Intn(len(candicate))]
```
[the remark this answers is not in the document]
CodingSinger: Thanks for the suggestion.

pkg/upstream/cluster/loadbalancer.go
CodingSinger: done.

taoyuanyuan: such as RoundRobinLB，we first choose host from a random index.
I think that we could do it here.
```
r := random
for i := r; i < len(hosts); i ++ {
}
for i := 0; i < r; i ++ {
}
```

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -129,5 +132,127 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+	// Check if the original host weights are equal and skip EDF creation if they are
+	if hostWeightsAreEqual(hosts) {
+		return
+	}
+
+	lb.scheduler = newEdfScheduler()
+
+	// Init Edf scheduler with healthy hosts.
+	for _, host := range hosts {
+		lb.scheduler.Add(host, lb.HostWight(host))
```
JasonRD, line 241: Where is the assignment for HostWight?

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -129,5 +132,127 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+	} else {
+		lb.choice = default_choice
+	}
+	lb.EdfLoadBalancer = newEdfLoadBalancerLoadBalancer(hosts, lb.unweightChooseHost)
+	return lb
+}
+
+func (lb *leastActiveRequestLoadBalancer) unweightChooseHost(context types.LoadBalancerContext) types.Host {
+
+	healthyHosts := lb.hosts.HealthyHosts()
```
wangfakang, line 156: Now `lb.hosts.HealthyHosts()` has been deprecated in this [PR](https://github.com/mosn/mosn/commit/d903bd10ececee4706605f01a73ec193ac9191bd#diff-8d8fb39636bed5042ff869294470b69c), So we should update the logical and you can refer [it](https://github.com/mosn/mosn/blob/master/pkg/upstream/cluster/loadbalancer.go#L72).

CodingSinger: done~

go.mod
```diff
@@ -35,7 +35,6 @@ require (
 	golang.org/x/net v0.0.0-20200222125558-5a598a2470a0
 	golang.org/x/sys v0.0.0-20190804053845-51ab0e2deafa
 	golang.org/x/text v0.3.2 // indirect
-	google.golang.org/appengine v1.4.0 // indirect
```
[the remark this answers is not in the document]
CodingSinger: sorry. my mistake.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -144,5 +147,133 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+		lb.mutex.Lock()
+		defer lb.mutex.Unlock()
+		var candicate types.Host
+		// Choose `choice` times and return the best one
+		// See The Power of Two Random Choices: A Survey of Techniques and Results
+		//  http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
+		for cur := 0; cur < int(lb.choice); cur++ {
+
+			randIdx := lb.rand.Intn(healthyHostLen)
+			tempHost := healthyHosts[randIdx]
```
wangfakang, line 193: we should be check the `tempHost.Health() `  health status.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -144,5 +147,146 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+	}
+	if total == 1 {
+		// Return directly if there is only one host
+		return targetHosts[0]
+	}
+	for i := 0; i < total; i++ {
+		if lb.scheduler != nil {
+			// do weight selection
+			host := lb.scheduler.Next().(types.Host)
+			lb.scheduler.Add(host, lb.hostWeightFunc(host))
```
wangfakang, line 229: When call `lb.scheduler.Add` function,  the value of  `edf.currentTime `  may already be another item.

pkg/upstream/cluster/edf.go
```diff
@@ -0,0 +1,75 @@
...
+// Pick entry with closest deadline and push again
+func (edf *edfSchduler) NextAndPush(weightFunc func(item WeightItem) float64) interface{} {
+	edf.lock.Lock()
+	defer edf.lock.Unlock()
+	if len(edf.items) == 0 {
+		return nil
+	}
+	entry := heap.Pop(&edf.items).(*edfEntry)
+	edf.currentTime = entry.deadline
+	weight := weightFunc(entry.item)
```
wangfakang, line 55: Need to deal with abnormal scenarios with a value of 0.

CodingSinger: add some comments for this func, it should be guaranteed by the person who provided the weight func method.

pkg/upstream/cluster/loadbalancer.go
CodingSinger: ok, i have mv this op to the lock block.

pkg/upstream/cluster/cluster.go
```diff
@@ -146,6 +146,8 @@ type clusterInfo struct {
 	lbOriDstInfo         types.LBOriDstInfo
 	tlsMng               types.TLSContextManager
 	connectTimeout       time.Duration
+	lbconfig	         v2.IsCluster_LbConfig
```
taoyuanyuan, line 149: gofmt

pkg/upstream/cluster/cluster.go
```diff
@@ -192,6 +194,10 @@ func (ci *clusterInfo) LbOriDstInfo() types.LBOriDstInfo {
 	return ci.lbOriDstInfo
 }
 
+func (ci *clusterInfo) LbConfig() v2.IsCluster_LbConfig{
```
taoyuanyuan, line 197: go fmt

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -19,6 +19,7 @@ package cluster
 
 import (
 	"math/rand"
+	v2 "mosn.io/mosn/pkg/config/v2"
```
taoyuanyuan, line 22: go imports

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -144,5 +147,146 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
+		return nil
+	}
+	if total == 1 {
+		// Return directly if there is only one host
+		return targetHosts[0]
+	}
+	for i := 0; i < total; i++ {
+		if lb.scheduler != nil {
+			// do weight selection
+			host := lb.scheduler.NextAndPush(lb.hostWeightFunc).(types.Host)
```
taoyuanyuan, line 229: host is unnecessary

pkg/upstream/cluster/edf.go
```diff
@@ -0,0 +1,76 @@
...
+func newEdfScheduler(cap int) *edfSchduler {
+	return &edfSchduler{
+		items: make(PriorityQueue, 0, cap),
+	}
+}
+
+// edfEntry is an internal wrapper for item that also stores weight and relative position in the queue.
+type edfEntry struct {
+	deadline float64
+	weight   float64
```
taoyuanyuan, line 23: weight not uesed ？

pkg/upstream/cluster/loadbalancer.go
CodingSinger: ok, i will delete it.

pkg/upstream/cluster/edf.go
CodingSinger: record the weight of entry.

PR #1074, opened 2020-03-30
Add a filter handler after chooseHost phase.

wangfakang:

In some scenarios, the filter needs to be executed after the `chooseHost`, such as modifying the request header based on the selected host.

### Solutions

Add a receiveFilter entry point `DownFilterAfterChooseHost`, which is executed after `ChooseHost ` and before `sendRequest`.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/proxy/downstream.go
```diff
@@ -627,7 +656,8 @@ func (s *downStream) getUpstreamProtocol() (currentProtocol types.ProtocolName)
 	return currentProtocol
 }
 
-func (s *downStream) receiveHeaders(endStream bool) {
+func (s *downStream) chooseHost(endStream bool) {
```
nejisama, line 659: choose host 为什么需要endStream作为参数？

pkg/proxy/downstream.go
```diff
@@ -1435,10 +1474,10 @@ func (s *downStream) processError(id uint32) (phase types.Phase, err error) {
 		return
 	}
 
-	if s.receiverFiltersAgain {
```
nejisama, line 1438: 这个逻辑修改是为什么

pkg/proxy/downstream.go
wangfakang: > choose host 为什么需要endStream作为参数？

在direct response场景下，此时会跳过`receiveHeaders`、`receiveData`，需要提前设置`s.downstreamRecvDone = endStream` 用于`endStream()`中做buffer复用处理。

pkg/proxy/downstream.go
wangfakang: > 这个逻辑修改是为什么

receiverFiltersAgain类型从`bool`调整为`types.Phase`类型，使其作为一个通用的重试阶段变量（当前已有的重试：route、ChooseHost）。

Issue #1088, opened 2020-04-05
Support Istio style of request factor match routing


trainyao:
### The New Feature

Fully support Istio style of request factor match routing, including tcp http routing

### Your scenes

MOSN Istio WG roadmap

### Your advice

Make Istio matching params work normally in MOSN

### Environment

- MOSN v0.10.0

wangfakang:
fixed.

PR #1089, opened 2020-04-05
bugfix(route): fix isito virtual host method matching not working

trainyao:

#1088 

### Solutions
check tcp http routing params work normally in MOSN
1. tcp routing params seems work normally in Istio 1.4.6, 1.5.1 is pending to be confirmed
1. found http method param not working normally in 1.4.6, fix in the first commit, other params is pending to be confirmed
1. maybe it's possible to use consistent hash to speed up matching route params, and maybe time complexity and space complexity should be considered and benchmark testing should be done when using consistent hash 

### UT result
on the way

### Benchmark
on the way

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/router/configutility.go
+++ b/pkg/router/configutility.go
@@ -18,6 +18,7 @@
 package router
 
 import (
+	"mosn.io/mosn/pkg/protocol"
 	"regexp"
 	"sort"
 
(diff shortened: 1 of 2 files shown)


pkg/router/configutility.go
```diff
@@ -42,6 +43,10 @@ func (cu *configUtility) MatchHeaders(requestHeaders api.HeaderMap, configHeader
 	for _, cfgHeaderData := range configHeaders {
 		cfgName := cfgHeaderData.Name.Get()
 		cfgValue := cfgHeaderData.Value
+		if cfgName == "method" {
```
wangfakang, line 46: Plz add a test for this fix, thx.

trainyao: unit-test is added

Issue #1071, opened 2020-03-27
Make MOSN support pilot-agent startup params


trainyao:
### The New Feature

Make MOSN support pilot-agent startup params, so pilot-agent is able to use MOSN as binary path to start

### Your scenes

MOSN Istio WG roadmap

#### advice 1 
MOSN use `urfave/cli` lib to manage cli command, it seems to support only subcommand format, but params from pilot-agent to startup envoy looks like:
```text
 /usr/local/bin/binaryPath -c /etc/istio/proxy/envoy-rev0.json --restart-epoch 0 --drain-time-s 45 --parent-shutdown-time-s 60 --service-cluster productpage.default --service-node sidecar~10.244.0.86~productpage-v1-7685fd594b-xql9q.default~default.svc.cluster.local --max-obj-name-len 189 --local-address-ip-version v4 --log-format [Envoy (Epoch 0)] [%Y-%m-%d %T.%e][%t][%l][%n] %v -l warning --component-log-level misc:error --concurrency 2
```
so to meet the need, maybe MOSN should change another cli manage lib, so start command be like
```text
mosn -c xxx ...
```
and maybe the `stop` `restart` subcommand in urfave/cli can change to be a flag, like:
```text
mosn --stop ...
mosn --restart ...
```
#### advice 2
I'm also thinking that the change can be pilot-agent side, like [this PR](https://github.com/mosn/istio/pull/5), pilot-agent can add a `proxtImplement` flag to choose which sidecar user can use, and mosn style startup params can be differ from envoy, in the pilot-agent
### Environment

- MOSN v0.10.0

wangfakang:
fixed.

PR #1102, opened 2020-04-11
feat(start): support istio startup params

trainyao:

#1071 

### Solutions
add some flags to receive params from istio, instead of throwing out 'flag not support error',
and also add logic to handle the 'log-level' flag to control log level in MOSN

### UT result
package cmd seems need no unit test? 

### Code Style
+ Show `Golint` result

### DIFF
--- a/cmd/mosn/main/control.go
+++ b/cmd/mosn/main/control.go
@@ -100,28 +100,28 @@ var (
 				EnvVar: "LOG_LEVEL",
 			}, cli.StringFlag{
 				Name:   "log-format, lf",
-				Usage:  "mosn log format",
+				Usage:  "mosn log format, currently useless",
 			}, cli.StringSliceFlag{
 				Name:  "component-log-level, lc",
-				Usage: "mosn component format",
+				Usage: "mosn component format, currently useless",
 			}, cli.StringFlag{
 				Name:  "local-address-ip-version",
-				Usage: "mosn log format",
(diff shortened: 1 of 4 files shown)


cmd/mosn/main/control.go
```diff
@@ -84,6 +94,34 @@ var (
...
+				EnvVar: "LOG_LEVEL",
+			}, cli.StringFlag{
+				Name:   "log-format, lf",
+				Usage:  "mosn log format",
+			}, cli.StringSliceFlag{
+				Name:  "component-log-level, lc",
+				Usage: "mosn component format",
+			}, cli.StringFlag{
+				Name:  "local-address-ip-version",
+				Usage: "mosn log format",
```
wangfakang, line 109: Usage description error, need to show that these parameters are not valid.

cmd/mosn/main/control.go
```diff
@@ -84,6 +94,34 @@ var (
...
+				Usage: "mosn log level",
+			}, cli.IntFlag{
+				Name:  "drain-time-s",
+				Usage: "mosn log format",
+			}, cli.StringFlag{
+				Name:  "parent-shutdown-time-s",
+				Usage: "mosn log format",
+			}, cli.IntFlag{
+				Name:  "max-obj-name-len",
+				Usage: "mosn log format",
```
wangfakang, line 121: Same question as above.

wangfakang:
LGTM.

Issue #1069, opened 2020-03-27
unify connection pool and hide from users


cch123:
### The New Feature

- A unified connection pool, and which can be depended on by the client library.
- An easy to use API user can directly init and use like `net/http` and other RPC frameworks.

### Your scenes

When developing our own protocols/features based on mosn,  we need to write a lot of code to implement a connection pool and its management which has nothing to do with business logic. eg. https://github.com/mosn/mosn/blob/master/test/lib/sofarpc/client.go#L27,  https://github.com/mosn/mosn/blob/master/test/lib/sofarpc/client.go#L182, https://github.com/mosn/mosn/blob/master/test/lib/sofarpc/client.go#L207, https://github.com/mosn/mosn/blob/master/test/lib/sofarpc/client.go#L86

### Your advice

- [ ] encapsulate the current internal API, provide easier user interface.

- [ ] unify connection pool implementations inside mosn(http/http2/xprotocol).

PR #1139, opened 2020-04-28
[WIP]Unify All Protocol Connection Pool

cch123:

https://github.com/mosn/mosn/issues/1069

### Solutions

- [x] put 3 conn pool logic together

- [x] support ping pong for xprotocol

- [ ]  extract stream from connpool pkg

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


cch123:
this pr will be divided to two prs

PR #1147, opened 2020-05-03
add index field to enhance edf scheduler

CodingSinger:
Add index field to enhance the edf scheduler. When the deadline is the same, the entry added first should has higher priority.

### DIFF
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -9,6 +9,7 @@ type edfSchduler struct {
 	lock        sync.Mutex
 	items       PriorityQueue
 	currentTime float64
+	globalIndex int64   // globalIndex records the current logical time
 }
 
 func newEdfScheduler(cap int) *edfSchduler {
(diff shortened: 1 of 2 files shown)


pkg/upstream/cluster/edf.go
```diff
@@ -54,16 +58,27 @@ func (edf *edfSchduler) NextAndPush(weightFunc func(item WeightItem) float64) in
 	entry := heap.Pop(&edf.items).(*edfEntry)
 	edf.currentTime = entry.deadline
 	weight := weightFunc(entry.item)
-	// update the deadline and put into priorityQueue again
+	// update the index、deadline and put into priorityQueue again
+	edf.globalIndex++
```
taoyuanyuan, line 62: maybe overflow ?

CodingSinger: I think it is difficult for int64 to overflow.

CodingSinger: or reset 0 when overflow?

taoyuanyuan: if reset 0, the result will be wrong.
use time.Now()，and compare time？

PR #1149, opened 2020-05-07
Feature strict dns cluster

dengqian:

### Solutions
Support resolve address in hosts to IP list, and apply load balance strategy based on IP list. 
Support update according to TTL, fixed duration.
Support configure dns resolve file instead of default /etc/resolv.conf. Also support configure dns resolve port.

configure example:

```
     {
        "Name": "testCluster",
        "type": "STRICT_DNS",
        "respect_dns_ttl": true,
        "lb_type": "LB_ROUNDROBIN",
        "max_request_per_conn": 1024,
        "conn_buffer_limit_bytes": 32768,
        "hosts": [
          {
            "address": "www.test.com:80"
          }
        ]
      }
```
### respect_dns_ttl (bool)
use TTL in dns result as refresh rate
### dns_refresh_rate (time, duration)
use fixed duration to update dns result
### dns_resolver_file (path, string)
use specified resolv.conf file instead of /etc/resolv.conf
### dns_resolver_port (port, string)
use specified port instead of default port 53
### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/upstream/cluster/strict_dns_cluster.go
```diff
@@ -0,0 +1,302 @@
...
+	// update current hosts in rt
+	rt.hosts = newHosts
+
+	// append hosts in other resolveTargets to newHosts
+	for _, r := range sdc.resolveTargets {
+		if r.config == rt.config {
+			continue
+		}
+		for _, h := range r.hosts {
+			newHosts = append(newHosts, h)
```
wangfakang, line 197: May cause the host  increase repeatedly, forexample
domainA: 1.1.1.1
domainB: 1.1.1.2

when domainA resolve response:
domainA：rt.hosts 1.1.1.1
domainB：rt.hosts 1.1.1.2
ALL host:
1.1.1.1
1.1.1.2

when domainB resolve response:
 domainA：rt.hosts 1.1.1.1 、1.1.1.2
 domainB：rt.hosts 1.1.1.2
ALL host:
1.1.1.2
1.1.1.1
1.1.1.2

pkg/config/v2/upstream.go
```diff
@@ -87,6 +94,11 @@ type Cluster struct {
 	Hosts                []Host              `json:"hosts,omitempty"`
 	ConnectTimeout       *api.DurationConfig `json:"connect_timeout,omitempty"`
 	LbConfig             IsCluster_LbConfig  `json:"lbconfig,omitempty"`
+	DnsRefreshRate       *api.DurationConfig `json:"dns_refresh_rate,omitempty"`
+	RespectDnsTTL        bool                `json:"respect_dns_ttl,omitempty"`
+	DnsLookupFamily      DnsLookupFamily     `json:"dns_lookup_family,omitempty"`
+	DnsResolverFile      string              `json:"dns_resolver_file,omitempty"`
```
wangfakang, line 100: Could you add a [dns_resolvers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto.html?highlight=strict%20dns) field. If the value of `dns_resolvers` field is empty, the dns resolver address specified by ` DnsResolverFile` defaultly.

pkg/upstream/cluster/strict_dns_cluster.go
dengqian: Later function hostSet.setFinalHost implements deduplication. So this place just leave it duplicated.

wangfakang: However, the resolution results of the current domain will append the resolution records of other domain.

pkg/config/v2/upstream.go
dengqian: done

pkg/upstream/cluster/strict_dns_cluster.go
```diff
@@ -0,0 +1,302 @@
...
+	defer sdc.mutex.Unlock()
+	sdc.StopResolve()
+
+	// before resolve, init hosts to origin unresolved address
+	sdc.simpleCluster.UpdateHosts(newHosts)
+
+	// update resolve targets and start new resolve task
+	var rts []*ResolveTarget
+	for _, host := range newHosts {
+		addr, port := getHostPortFromAddr(host.AddressString())
```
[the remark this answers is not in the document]
dengqian: fixed

pkg/upstream/cluster/strict_dns_cluster.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+		rt.resolveTimer.Stop()
+		rt.resolveTimeout.Stop()
+	}()
+
+	// start resolve now
+	rt.resolveTimer = utils.NewTimer(0, rt.OnResolve)
+	for {
+		select {
+		case <-rt.stop:
```
wangfakang, line 244: Should be call `rt.resolveTimer.Stop()` , Because `UpdateClusterHosts()` and `OnResolve()` may be called by multiple goroutines, this will cause overwriting problems with data updates.  In addition, we should give priority to the `rt.stop` event, forexample：
```
for {
    select {
    case <-rt.stop:
        xxx
    default:
        select {
        case <-rt.stop:
            xxx
        case rt.timeout:
           xxx
        }
    }
}
```

pkg/upstream/cluster/strict_dns_cluster.go
```diff
@@ -0,0 +1,311 @@
...
+		rt := &ResolveTarget{
+			dnsAddress:       addr,
+			port:             port,
+			config:           &config,
+			refreshTimeout:   DefaultRefreshTimeout,
+			dnsRefreshRate:   make(chan time.Duration),
+			stop:             make(chan struct{}),
+			timeout:          make(chan bool),
+			strictDnsCluster: sdc,
+			hosts:            []types.Host{host},
```
wangfakang, line 134: Maybe we need to add a `version` field to `ResolveTarget` struct to determine whether it is the same as the cluster version. If they are inconsistent, there is no need to call `sdc.simpleCluster.UpdateHosts (newHosts)` in `updateDynamicHosts`. forexample：

```
func (sdc *strictDnsCluster) UpdateHosts(newHosts []types.Host) {  
    ……
    atomic.AddUint64(&sdc.version , 1)
   ……
}

func (sdc *strictDnsCluster) updateDynamicHosts(newHosts []types.Host, rt *ResolveTarget) {
……
v := atomic.LoadUint64(&sdc.version)
// don't overwriting 
if rt.version != v {
    continue
}
……

}

```

PR #1161, opened 2020-05-15
use edf to imlement wrr

CodingSinger:
the time complexity of edf schedule wrr is O (logn), better than O (n) of smoothWeightedRRLB.

### DIFF
--- a/pkg/upstream/cluster/loadbalancer.go
+++ b/pkg/upstream/cluster/loadbalancer.go
@@ -106,6 +106,7 @@ func (lb *randomLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) int {
 type roundRobinLoadBalancer struct {
 	hosts   types.HostSet
 	rrIndex uint32
+	*EdfLoadBalancer
 }
 
 type roundRobinLoadBalancerFactory struct {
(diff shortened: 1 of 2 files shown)


pkg/upstream/cluster/loadbalancer.go
```diff
@@ -121,13 +122,20 @@ func (f *roundRobinLoadBalancerFactory) newRoundRobinLoadBalancer(info types.Clu
...
+	return rrLB
+}
+
+func (lb *roundRobinLoadBalancer) hostWeight(item WeightItem) float64 {
+	host := item.(types.Host)
+	return float64(host.Weight())
 }
 
-func (lb *roundRobinLoadBalancer) ChooseHost(context types.LoadBalancerContext) types.Host {
+func (lb *roundRobinLoadBalancer) unweightChooseHost(context types.LoadBalancerContext) types.Host {
```
wangfakang, line 138: There is no need to call `host.Health()`  in `unweightChooseHost()`  function.

CodingSinger: yes. Already judged in `ChooseHost`

CodingSinger:
> do not modify the simple round robin load balance.
> you can use the edf lb instead of wrr lb.

ok, i will to change it.

PR #1173, opened 2020-05-28
use edf wrr to cover the original smooth wrr

CodingSinger:
the time complexity of edf schedule wrr is O (logn), better than O (n) of smoothWeightedRR LB.
Relate #1161

### DIFF
--- a/pkg/upstream/cluster/loadbalancer.go
+++ b/pkg/upstream/cluster/loadbalancer.go
@@ -46,7 +46,7 @@ func init() {
 	}
 	RegisterLBType(types.RoundRobin, rrFactory.newRoundRobinLoadBalancer)
 	RegisterLBType(types.Random, newRandomLoadBalancer)
-	RegisterLBType(types.WeightedRoundRobin, newSmoothWeightedRRLoadBalancer)
+	RegisterLBType(types.WeightedRoundRobin, newWeightedRRLoadBalancer)
 	RegisterLBType(types.LeastActiveRequest, newleastActiveRequestLoadBalancer)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/upstream/cluster/loadbalancer.go
```diff
@@ -152,80 +152,49 @@ func (lb *roundRobinLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) in
...
 }
 
+func (lb *WeightedRRLoadBalancer) hostWeight(item WeightItem) float64 {
+	host := item.(types.Host)
+	return float64(host.Weight())
+}
+// do unweighted (fast) selection
+func (lb *WeightedRRLoadBalancer) unweightChooseHost(context types.LoadBalancerContext) types.Host {
+	targets := lb.hosts.Hosts()
+	total := len(targets)
```
wangfakang, line 193: need check value of `total`, if it's 0 should return nil.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -46,7 +46,7 @@ func init() {
 	}
 	RegisterLBType(types.RoundRobin, rrFactory.newRoundRobinLoadBalancer)
 	RegisterLBType(types.Random, newRandomLoadBalancer)
-	RegisterLBType(types.WeightedRoundRobin, newSmoothWeightedRRLoadBalancer)
+	RegisterLBType(types.WeightedRoundRobin, newWeightedRRLoadBalancer)
```
wangfakang, line 49: Maybe better that  named `newWeightedWRRLoadBalancer` .

pkg/upstream/cluster/loadbalancer.go
CodingSinger: Already judged in the `EdfLoadbalancer#ChooseHost`.

pkg/upstream/cluster/loadbalancer.go
CodingSinger: I think WeigthtedRR = WRR?

wangfakang: ok

PR #1180, opened 2020-05-29
fix: change sentinel log path

ansiz:

### Solutions

change sentinel log path to fix panic due to sentinel path permission problem

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/filter/stream/flowcontrol/stream_filter_factory.go
+++ b/pkg/filter/stream/flowcontrol/stream_filter_factory.go
@@ -4,18 +4,35 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"os"
+	"sync"
 
-	"mosn.io/mosn/pkg/types"
-
+	sentinel "github.com/alibaba/sentinel-golang/api"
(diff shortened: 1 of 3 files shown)


pkg/filter/stream/flowcontrol/stream_filter_factory.go
```diff
@@ -73,6 +101,9 @@ func createRpcFlowControlFilterFactory(conf map[string]interface{}) (api.StreamF
 		log.DefaultLogger.Errorf("update rules failed")
 		return nil, err
 	}
+	initOnce.Do(func() {
+		initSentinel(flowControlCfg.AppName, flowControlCfg.LogPath)
```
wangfakang, line 105: Maybe TODO
support dynamicly update config for initSentinel.

ansiz: done

ansiz: > Maybe TODO
> 
> support dynamicly update config for initSentinel.

done

PR #1181, opened 2020-05-30
updated dockerfile

ducksecops:

Updated Ubuntu in dockerfile to bionic (18.04) current trusty (14.04) was end of life April 2019.
Updated golang in dockerfie to 1.13 branch as current 1.12 branch is now eol with release of 1.14


nejisama:
Upgrading golang version requires some adaptation work, this is is in our plan, but not now

PR #1183, opened 2020-06-01
Feature istio1.6 adapter

champly:
adapt istio 1.6

- compatible `envoy.listener.original_dst` listener filter
- log format variable transform lower, if not support use raw string


champly:
@wangfakang PTAL

PR #1187, opened 2020-06-02
support pre stop hook.

zonghaishang:

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"mosn.io/mosn/pkg/proxy"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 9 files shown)


zonghaishang:
@cch123 @nejisama please take a look

pkg/server/handler.go
```diff
@@ -488,8 +489,91 @@ func (al *activeListener) OnNewConnection(ctx context.Context, conn api.Connecti
...
+			}
+		}
+	}
+
+	return activeStream
+}
+
 func (al *activeListener) OnClose() {}
 
+func (al *activeListener) PreStopHook(ctx context.Context) func() error {
```
cch123, line 536: explain why this function is needed and when to use it

Issue #1194, opened 2020-06-03
Optimize log printing


wenxuwan:
### Describe the bug
The log prints of phase is not good for read, eg:

![image](https://user-images.githubusercontent.com/11661906/83642223-1febe780-a5e1-11ea-84ab-343eef15323f.png)

if i want to know what's 4 and 6 i need read the code, and the "enum" type is hard to read.

May be like this is better:

![image](https://user-images.githubusercontent.com/11661906/83646303-0c8f4b00-a5e6-11ea-8f53-c8d97eede3d7.png)

### Environment

- MOSN Version

PR #1195, opened 2020-06-04
Optimize log printing

wenxuwan:
The content of phase prints is hard to read, so add the phase name for easy read.

Fixes gh-1194

#1194 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/types/proxy.go
+++ b/pkg/types/proxy.go
@@ -22,7 +22,26 @@ import (
 )
 
 var (
-	ErrExit = errors.New("downstream process completed")
+	ErrExit   = errors.New("downstream process completed")
+	PhaseName = []string{
+		InitPhase:                 "InitPhase",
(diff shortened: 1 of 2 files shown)


pkg/types/proxy.go
```diff
@@ -22,7 +22,8 @@ import (
 )
 
 var (
-	ErrExit = errors.New("downstream process completed")
+	ErrExit   = errors.New("downstream process completed")
```
taoyuanyuan, line 25: it is easy to make mistakes when add new phase.
could you please optimize it ?
e.g.
```
var PhaseName = []string{
	InitPhase:  "InitPhase",
	DownFilter: "DownFilter",
	MatchRoute: "MatchRoute",
	End:        "End",
}
```

wenxuwan: done.

PR #1200, opened 2020-06-07
adapte istio pilot-agent readiness check & xDS requets's node info with pod labels

champly:
- Istio pilot-agent readiness check add `listener_manager.workers_started` from `http://127.0.0.1:15000/stats`, reference [code](https://github.com/istio/istio/blob/master/pilot/cmd/pilot-agent/status/util/stats.go#L86)
- #1208


pkg/admin/server/istio.go
```diff
@@ -13,15 +13,16 @@ import (
...
+	CDS_UPDATE_REJECT    = "cluster_manager.cds.update_rejected"
+	LDS_UPDATE_SUCCESS   = "listener_manager.lds.update_success"
+	LDS_UPDATE_REJECT    = "listener_manager.lds.update_rejected"
+	SERVER_STATE         = "server.state"
+	STAT_WORKERS_STARTED = "listener_manager.workers_started"
 )
 
 func statsForIstio(w http.ResponseWriter, r *http.Request) {
-	state, err := getIstioState()
+	state, workersStarted, err := getIstioState()
```
wangfakang, line 25: set the value of `workersStarted` could  via state.

champly:
Add xDS request node info with pod labels info

pkg/admin/server/istio.go
champly: already reset getIstioState method

PR #1212, opened 2020-06-10
bugfix: http2 connect before creating stream client

dunjut:

The `newActiveClient` function inside the http2 package [creates stream client before setting up a real connection to the remote](https://github.com/mosn/mosn/blob/master/pkg/stream/http2/connpool.go#L212-L219), which leads to [nil pointer dereference of connection.rawConnection](https://github.com/mosn/mosn/blob/master/pkg/network/connection.go#L560) because creating stream client would [write http2 preface](https://github.com/mosn/mosn/blob/master/pkg/module/http2/mhttp2.go#L1022) and settings to connection (though the underlying connection is not ready yet).

### Solutions

Move the `Connect()` code in front of `pool.createStreamClient()`.

### DIFF
--- a/pkg/stream/http2/connpool.go
+++ b/pkg/stream/http2/connpool.go
@@ -156,7 +156,6 @@ func (p *connPool) onConnectionEvent(client *activeClient, event api.ConnectionE
 	} else if event == api.ConnectTimeout {
 		host.HostStats().UpstreamRequestTimeout.Inc(1)
 		host.ClusterInfo().Stats().UpstreamRequestTimeout.Inc(1)
-		client.client.Close()
 	} else if event == api.ConnectFailed {
 		host.HostStats().UpstreamConnectionConFail.Inc(1)
 		host.ClusterInfo().Stats().UpstreamConnectionConFail.Inc(1)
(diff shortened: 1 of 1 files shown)


pkg/stream/http2/connpool.go
```diff
@@ -208,6 +208,11 @@ func newActiveClient(ctx context.Context, pool *connPool) *activeClient {
 
 	host := pool.Host()
 	data := host.CreateConnection(ctx)
+	if err := data.Connection.Connect(); err != nil {
+		log.DefaultLogger.Debugf("http2 underlying connection error: %v", err)
+		return nil
+	}
+
```
cch123, line 215: 这样似乎会影响后面 activeClient.OnEvent 里的一些异常统计吧

cch123: codecClient.AddConnectionEventListener(ac)
=> 
data.Connection.AddConnectionEventListener(ac)

这样应该就可以了，这个修改点，我也顺便合并到之后统一的连接池里吧

dunjut:
cc @taoyuanyuan PTAL

PR #1223, opened 2020-06-22
avoid reentrance of Stop for same heartbeat object

cch123:

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/stream/connpool/msgconnpool/activeclient.go
```diff
@@ -69,8 +69,12 @@ func (ac *activeClient) setHeartBeater(kp KeepAlive) {
 func (ac *activeClient) clearHeartBeater() {
 	// clear the previous keepAlive
 	if ac.keepAlive != nil && ac.keepAlive.keepAlive != nil {
```
taoyuanyuan, line 71: maybe panic.

PR #1249, opened 2020-07-02
Optimize and adjust CEL

wzshiming:

### Solutions
- CEL supports simple registration functions
- Definition of separation mixer extraction function
- Calculate request size on demand

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/filter/stream/stats/config.go
+++ b/pkg/filter/stream/stats/config.go
@@ -19,8 +19,6 @@ package stats
 
 import (
 	"time"
-
-	"mosn.io/mosn/pkg/cel/attribute"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/filter/stream/stats/config.go
```diff
@@ -0,0 +1,126 @@
...
+package stats
+
+import (
+	"time"
+)
+
+// https://istio.io/latest/docs/reference/config/proxy_extensions/stats/#MetricDefinition
+type StatsConfig struct {
+	// The following settings should be rarely used. Enable debug for this filter.
+	Debug string `json:"debug"`
```
wangfakang, line 27: code style.

PR #1261, opened 2020-07-09
feat: http redirect

### LINKED ISSUE #1255: Feature Request: http redirect
### The New Feature

Currently MOSN is lack of the support of http redirect, which makes user unable to perform redirect in [`VirtualService`](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRedirect).

It is worth noting that envoy has this feature https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-redirectaction, and I noticed that related code was commented out in MOSN: https://github.com/mosn/mosn/blob/0da49757f0cc75955813f3372d3ff5ce4ce1cf7c/pkg/xds/conv/convertxds.go#L1063-L1074

~~Is it intended that MOSN will not have this feature or is it just not implemented yet?~~

### Roadmap

* [ ] add `RedirectResponseRule` interface to `mosn.io/api` https://github.com/mosn/api/pull/10
* [ ] add `RedirectAction` struct to `mosn/pkg/config/v2` and implement `RedirectResponseRule` interface in `mosn/pkg/router`
* [ ] convert `xdsroute.RedirectAction` to redirect action that mosn is able to recognize

knight42:

Fixes #1255 

### Solutions

send redirect response to client, similar to how we handle direct response.

NOTE: this PR relies on https://github.com/mosn/api/pull/10

TODO:
* [x] add integration test

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


knight42:
cc @wangfakang

pkg/config/v2/route.go
```diff
@@ -280,6 +281,13 @@ type RouterMatch struct {
 	Headers []HeaderMatcher `json:"headers,omitempty"` // Match request's Headers
 }
 
+// RedirectAction represents the redirect response parameters
+type RedirectAction struct {
```
wangfakang, line 285: Need support `scheme_redirect`，forexample
`http://mosn.io`  -> `https://mosn.io` or `https://mosn.io`  -> `http://mosn.io` 

refer: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-redirectaction-scheme-redirect

knight42: @wangfakang  Got it, but how could we know the current scheme? Perhaps I have missed something, I am unable to find any related headers or methods to get the current scheme.

wangfakang: > @wangfakang Got it, but how could we know the current scheme? Perhaps I have missed something, I am unable to find any related headers or methods to get the current scheme.

Could add a 'scheme' variable to get current request `scheme`, here is a var [refer](https://github.com/mosn/mosn/blob/master/pkg/stream/http/var.go).

pkg/stream/http2/stream.go
```diff
@@ -299,6 +299,7 @@ func (conn *serverStreamConnection) handleFrame(ctx context.Context, i interface
 		URL, _ := url.Parse(URI)
 		h2s.Request.URL = URL
 
+		header.Set(protocol.MosnHeaderSchemeKey, scheme)
```
wangfakang, line 302: Need del `MosnHeaderSchemeKey` header at the `(s *clientStream) AppendHeaders ` function.

pkg/router/base_rule.go
```diff
@@ -19,12 +19,15 @@ package router
...
 	"context"
+	"fmt"
 	"math/rand"
+	"net/http"
 	"strings"
 	"sync"
 	"time"
 
 	"mosn.io/api"
+
```
wangfakang, line 30: import format.

pkg/router/base_rule.go
```diff
@@ -122,13 +127,41 @@ func NewRouteRuleImplBase(vHost *VirtualHostImpl, route *v2.Router) (*RouteRuleI
...
+	}
 	return base, nil
 }
 
 func (rri *RouteRuleImplBase) DirectResponseRule() api.DirectResponseRule {
 	return rri.directResponseRule
 }
 
+func (rri *RouteRuleImplBase) RedirectResponseRule() api.RedirectResponseRule {
+	if rri.redirectResponseRule == nil {
```
wangfakang, line 159: It looks like it's equivalent to just return `rri.redirectResponseRule` ?

pkg/router/base_rule.go
knight42: hmm this is how my goland format the code..

pkg/router/base_rule.go
knight42: actually they are subtly different. if `rri.redirectResponseRule` is nil and we simply return `rri.redirectResponseRule` here, `RedirectResponseRule() == (*redirectResponseImpl)(nil)`  rather than `nil`, and we have to call `reflect.ValueOf(..).IsNil()` at https://github.com/mosn/mosn/pull/1261/files#diff-ba11a00bb14720d45706e6dee657a6aeR713 to check if it is actually `nil`.

knight42: you could check the example here https://play.golang.org/p/y-ZgiSzIo1r

wangfakang: Cool, you are right.

pkg/stream/http2/stream.go
```diff
@@ -435,13 +436,16 @@ type serverStream struct {
...
 
+	// --- do not pass internal headers to upstream
 	var status int
 	if value, _ := headers.Get(types.HeaderStatus); value != "" {
 		headers.Del(types.HeaderStatus)
 		status, _ = strconv.Atoi(value)
 	} else {
 		status = 200
 	}
+	headers.Del(types.HeaderScheme)
```
wangfakang, line 447: It should be call it in the function of client's [AppendHeaders](https://github.com/mosn/mosn/pull/1261/files#diff-70118d13dd7afdc9f6b24253246ae52fR805) rather than serverStream.

pkg/stream/http2/stream.go
```diff
@@ -299,6 +299,7 @@ func (conn *serverStreamConnection) handleFrame(ctx context.Context, i interface
 		URL, _ := url.Parse(URI)
 		h2s.Request.URL = URL
 
+		header.Set(types.HeaderScheme, scheme)
```
wangfakang, line 302: maybe named `protocol.MosnHeaderScheme` would be better.

pkg/xds/conv/convertxds.go
```diff
@@ -1060,18 +1060,16 @@ func convertRetryPolicy(xdsRetryPolicy *xdsroute.RetryPolicy) *v2.RetryPolicy {
...
- }
-*/
+func convertRedirectAction(xdsRedirectAction *xdsroute.RedirectAction) *v2.RedirectAction {
+	if xdsRedirectAction == nil {
+		return nil
+	}
+	return &v2.RedirectAction{
+		HostRedirect: xdsRedirectAction.GetHostRedirect(),
+		PathRedirect: xdsRedirectAction.GetPathRedirect(),
+		ResponseCode: int(xdsRedirectAction.GetResponseCode()),
```
wangfakang, line 1070: Need convert `SchemeRedirect` value via `GetSchemeRedirect`.

pkg/router/base_rule.go
```diff
@@ -122,13 +127,41 @@ func NewRouteRuleImplBase(vHost *VirtualHostImpl, route *v2.Router) (*RouteRuleI
 			body:   route.DirectResponse.Body,
 		}
 	}
+	// add redirect repsonse rule
+	if route.Redirect != nil {
+		r := route.Redirect
+		rule := &redirectImpl{
+			path:   r.PathRedirect,
+			host:   r.HostRedirect,
+			scheme: r.SchemeRedirect,
```
wangfakang, line 136: Should be check the legality of scheme.

pkg/proxy/downstream.go
```diff
@@ -697,6 +709,41 @@ func (s *downStream) chooseHost(endStream bool) {
 		}
 		return
 	}
+
+	if rule := s.route.RedirectRule(); rule != nil {
```
wangfakang, line 713: If both the `DirectResponseRule` and `RedirectRule` rules are configured,  it would be better that `RedirectRule` rule had a higher priority?

pkg/router/base_rule.go
```diff
@@ -33,6 +37,10 @@ import (
 	"mosn.io/mosn/pkg/upstream/cluster"
 )
 
+var (
+	schemeValidator = regexp.MustCompile("^[a-z][a-z0-9.+-]*$")
```
wangfakang, line 41: You can annotate the RFC referenced below.

knight42: added

pkg/stream/http2/stream.go
```diff
@@ -802,6 +803,8 @@ func (s *clientStream) AppendHeaders(ctx context.Context, headersIn api.HeaderMa
 	var req *http.Request
 	var isReqHeader bool
 
+	headersIn.Del(protocol.MosnHeaderScheme)
```
wangfakang, line 806: `headersIn.Del(protocol.MosnHeaderScheme)` should be called after `headersIn = headersIn.Clone()`.

knight42: nice catch! fixed

PR #1267, opened 2020-07-14
mtls support hash and update

nejisama:
## tls context
+ Add a new type `HashValue`  to represent tls config in tls context.
   + How to generate HashValue can be extended in `confighook`
+ Update tls config is valid  for sds mode
+ Parse client auth in tls config can be extended in `confighook`

## connection pool
+ connpool changed when tls hash value changed

## cluster manager
+ Add a new api `GetAllClusters`
+ Update cluster config will update the hosts' cluster info 

## benchmark
+ Add update cluster info 
```
BenchmarkAddOrUpdateCluster/update_count:100-4              10000      181294 ns/op     30344 B/op       115 allocs/op
BenchmarkAddOrUpdateCluster/update_count:500-4               5000      409458 ns/op    100307 B/op       127 allocs/op
BenchmarkAddOrUpdateCluster/update_count:1000-4              2000      703845 ns/op    184069 B/op       141 allocs/op
BenchmarkAddOrUpdateCluster/update_count:5000-4               500     2797682 ns/op    768478 B/op       219 allocs/op
BenchmarkAddOrUpdateCluster/update_count:20000-4              100    12816462 ns/op   2974942 B/op       582 allocs/op
```
+ Before

```
BenchmarkAddOrUpdateCluster/update_count:100-4         	   10000	    210924 ns/op	   30343 B/op	     115 allocs/op
BenchmarkAddOrUpdateCluster/update_count:500-4         	    3000	    415557 ns/op	  100287 B/op	     127 allocs/op
BenchmarkAddOrUpdateCluster/update_count:1000-4        	    3000	    663567 ns/op	  184042 B/op	     140 allocs/op
BenchmarkAddOrUpdateCluster/update_count:5000-4        	     500	   2711602 ns/op	  768509 B/op	     219 allocs/op
BenchmarkAddOrUpdateCluster/update_count:20000-4       	     100	  12564153 ns/op	 2974319 B/op	     577 allocs/op
```

### DIFF
--- a/pkg/mtls/ciphers.go
+++ b/pkg/mtls/ciphers.go
@@ -1,0 +1,58 @@
+// +build !BabaSSL
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package mtls
+
+import (
+	"mosn.io/mosn/pkg/mtls/crypto/tls"
+)
+
+// Ciphers
+var (
+	defaultCiphers = []uint16{
+		tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+		tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+		tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+		tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+		tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+		tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
+		tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
+		tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+		tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
+		tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_128_CBC_SHA,
+	}
+	ciphersMap = map[string]uint16{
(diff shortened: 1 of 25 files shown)


pkg/mtls/sds/subscriber.go
```diff
@@ -126,7 +126,7 @@ func (subscribe *SdsSubscriber) sendRequestLoop(sdsStreamClient *SdsStreamClient
 			for {
 				err := subscribe.sendRequest(discoveryReq)
 				if err != nil {
-					log.DefaultLogger.Alertf("sds.subscribe", "[xds] [sds subscriber] send sds request fail , resource name = %v", name)
+					log.DefaultLogger.Alertf("sds.subscribe.request", "[xds] [sds subscriber] send sds request fail , resource name = %v", name)
```
wangfakang, line 129: should be used `log.DefaultLogger.Alertf("sds.subscribe.request [xds] [sds subscriber] send sds request fail , resource name = %v", name)`  instead of  `log.DefaultLogger.Alertf("sds.subscribe.request", "[xds] [sds subscriber] send sds request fail , resource name = %v", name)`, otherwise there will be logging "xx !(EXTRA string= xx"

pkg/mtls/sds/subscriber.go
```diff
@@ -216,7 +216,7 @@ func (subscribe *SdsSubscriber) reconnect() {
 	for {
 		sdsStreamConfig, err := subscribe.convertSdsConfig(subscribe.sdsConfig)
 		if err != nil {
-			log.DefaultLogger.Alertf("sds.subscribe", "[xds][sds subscriber] convert sds config fail %v", err)
+			log.DefaultLogger.Alertf("sds.subscribe.config", "[xds][sds subscriber] convert sds config fail %v", err)
```
wangfakang, line 219: same problem as above.

pkg/mtls/ciphers.go
```diff
@@ -0,0 +1,58 @@
+// +build !BabaSSL
```
wangfakang, line 1: Need `// +build !BabaSSL` ?

pkg/mtls/sds/subscriber.go
```diff
@@ -60,13 +60,13 @@ func (subscribe *SdsSubscriber) Start() {
 	for {
 		sdsStreamConfig, err := subscribe.convertSdsConfig(subscribe.sdsConfig)
 		if err != nil {
-			log.DefaultLogger.Alertf("sds.subscribe", "[sds][subscribe] convert sds config fail %v", err)
+			log.DefaultLogger.Alertf("sds.subscribe.config", "[sds][subscribe] convert sds config fail %v", err)
```
wangfakang, line 63: Same problem as above.

pkg/mtls/ciphers.go
william-zk: > Need `// +build !BabaSSL` ?
go build tag="" equal to go build tags=!BabaSSL

pkg/mtls/sds/subscriber.go
nejisama: no, alertf is different

pkg/mtls/ciphers.go
william-zk: > Need `// +build !BabaSSL` ?

I'm not sure what your mean, if you means that whether should go build tags=xxx, see this

> > Need `// +build !BabaSSL` ?
> > go build tag="" equal to go build tags=!BabaSSL

if you means that whether need this code "// +build !BabaSSL"
you can have a look at file cipher_cgo.go, we use build tag to implement feature distinguish

PR #1270, opened 2020-07-16
Bugfix: prefix variable in access log is not working

dengqian:
prefix variables such as request_header_xxx in access log is not parsed correctly, variable names will always only save the prefix

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/log/accesslog_test.go
+++ b/pkg/log/accesslog_test.go
@@ -281,11 +281,15 @@ func prepareLocalIpv4Ctx() context.Context {
 		"service": "test",
 	}
 	ctx = context.WithValue(ctx, requestHeaderMapKey, reqHeaders)
+	reqHeaderMap := newHeaderMap(reqHeaders)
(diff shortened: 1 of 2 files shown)


pkg/log/accesslog_test.go
```diff
@@ -654,25 +685,25 @@ func upstreamHostGetter(ctx context.Context, value *variable.IndexedValue, data
 }
 
 func requestHeaderMapGetter(ctx context.Context, value *variable.IndexedValue, data interface{}) (string, error) {
-	headers := ctx.Value(requestHeaderMapKey).(api.HeaderMap)
+	headers := ctx.Value(requestHeaderMapKey).(map[string]string)
```
wangfakang, line 688: Maybe it would be better to modify `prepareLocalIpv4Ctx()` and `prepareLocalIpv6Ctx`.

PR #1293, opened 2020-07-30
remove redundant 'replace' directive in go.mod

### LINKED ISSUE #1292: No need to pin github.com/envoyproxy/go-control-plane to v0.9.4 twice
The `go-control-plane`'s version is specified twice in go.mod (but the value is the same):
https://github.com/mosn/mosn/blob/5bc5c6551b16aa2e65ae7e2ae559fc7e376aa20d/go.mod#L13
https://github.com/mosn/mosn/blob/5bc5c6551b16aa2e65ae7e2ae559fc7e376aa20d/go.mod#L54

AFAIK, we only need to specify the version once.

spacewander:
Since now we already use go-control-plane v0.9.4, there is no need
to apply the replace directive.

Close #1292.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,5 +50,3 @@ require (
 	mosn.io/api v0.0.0-20200720094309-63de04640d4e
 	mosn.io/pkg v0.0.0-20200723115157-3842674e37c7
 )
-
-replace github.com/envoyproxy/go-control-plane => github.com/envoyproxy/go-control-plane v0.9.4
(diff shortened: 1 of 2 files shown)


cch123:
revert reason
go-control-plane breaks its API every time it updates, we need to have a stable build and update this library at our own pace, so reverted

spacewander:
@cch123 
It's strange. The version of go-control-plane is already pinned in the build list even without the replace directive. Go mod should not update it.

According to the doc of go.mod's `replace`:
> The same module version must not also appear in the build list.

https://golang.org/ref/mod#tmp_15

cch123:
@spacewander, if any other lib was imported and also depends on go-control-plane, go mod will try to update your dependencies according to semver rules, eg, go-control-plane : 0.9.4 -> 0.9.8, which probably breaks builds.

the dependency list cannot lock your dependency version.

and I tries to only reserve this rule only in replace block, but when execute go mod vendor, go mod will automatically add go-control-plane to require block

spacewander:
@cch123 
I see. The version of go-control-plane dependency list in require{} block is equal to `max(your requirement, your dependencies' requirement).` And the result happens to be `0.9.4` in the latest commit, which is also the version for your requirement.

In the future, one of {mosn, mosn dependencies} may upgrade their required go-control-plane's version, and `go-control-plane` doesn't keep the backward compatibility, therefore you control to pin the version with `replace` directive.

PR #1318, opened 2020-08-18
bugfix: UDP session read buffer size should not be reset on read timeout

dengqian:

c.readBuffer will be reset on read timeout,  which will resize UDP recv buffer to 128. Data in a UDP packet exceeding the buffer size will be discarded.

### Solutions
c.readBuffer of a UDP session should not be reset.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/mosn/starter.go
+++ b/pkg/mosn/starter.go
@@ -225,6 +225,13 @@ func (m *Mosn) beforeStart() {
 			ln.Close()
 		}
 	}
+	for _, ln := range m.inheritPacketConn {
+		if ln != nil {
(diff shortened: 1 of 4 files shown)


pkg/mosn/starter.go
```diff
@@ -225,6 +225,12 @@ func (m *Mosn) beforeStart() {
 			ln.Close()
 		}
 	}
+	for _, ln := range m.inheritPacketConn {
```
wangfakang, line 228: It would be better to add a comment.

taoyuanyuan (approved):
LGTM

PR #1319, opened 2020-08-20
fix data race problem

wenxuwan:

image

from the picture we can find the curClientStreams value change to -1, so here change to atomic operation.

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/module/http2/mhttp2.go
+++ b/pkg/module/http2/mhttp2.go
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"golang.org/x/net/http/httpguts"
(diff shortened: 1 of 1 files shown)


pkg/module/http2/mhttp2.go
```diff
@@ -711,9 +712,9 @@ func (sc *MServerConn) closeStream(st *stream, err error) {
 	if sc.delStream(st.id) {
 		st.state = stateClosed
 		if st.isPushed() {
-			sc.curPushedStreams--
+			atomic.AddUint32(&sc.curPushedStreams, ^uint32(-(-1)-1))
```
taoyuanyuan, line 715: ^uint32(0)

wenxuwan: done

taoyuanyuan (approved):
LGTM

PR #1322, opened 2020-08-21
bugfix: status code not set after retry

dengqian:

UpstreamRequest.protocol is nil in retry scenario, which is needed to get status code from headers in protocol.MappingHeaderStatusCode(r.downStream.context, r.protocol, headers)

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


taoyuanyuan (approved):
LGTM

PR #1352, opened 2020-09-08
Config extend

nejisama:
1.  Move registry config to extend. different registry implementation can have different registry config
2. Support Extend configs to auto dump.
3. All configs changed will be dumped into config file if auto config is enabled. 
    + merge admin/store/effectiveconfig into configmanager
4. New test framework test cases add to ci
5. Add a debug package for test case. default build will no such functions registered.


pkg/admin/server/apis.go
```diff
@@ -81,48 +82,57 @@ func configDump(w http.ResponseWriter, r *http.Request) {
...
 		case "router":
-			v := store.GetMOSNConfig(store.CfgTypeRouter)
-			routerInfo, ok := v.(map[string]v2.RouterConfiguration)
-			if ok && len(param) > 0 {
-				info = routerInfo[param[0]]
-			}
+			configmanager.HandleMOSNConfig(configmanager.CfgTypeRouter, func(v interface{}) {
+				routerInfo, ok := v.(map[string]v2.RouterConfiguration)
+				if ok && len(param) > 0 {
+					handle(routerInfo[param[0]])
```
cch123, line 112: The function literal capture the loop variable, and always use the last param of  r.Form?

taoyuanyuan (approved):
LGTM

PR #1371, opened 2020-09-17
Optimize router regex config.

wangfakang:
Optimize the Router configuration to avoid dumps useless config when no RegexRewrite.


nejisama (approved):
lgtm

PR #1376, opened 2020-09-21
support multiple connections on multiplex conn pool

cch123:

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


cch123:
use dubbo to test:

pkg/stream/xprotocol/connpool_multiplex.go
```diff
@@ -59,24 +69,30 @@ func (p *poolMultiplex) init(client *activeClientMultiplex, sub types.ProtocolNa
...
-			p.activeClients.Delete(sub)
+			p.activeClients[index].Delete(sub)
 		}
 	}, nil)
 }
 
 // CheckAndInit init the connection pool
 func (p *poolMultiplex) CheckAndInit(ctx context.Context) bool {
 	var client *activeClientMultiplex
+	clientIdx := atomic.AddInt64(&p.currentCheckAndInitIdx, 1) % int64(len(p.activeClients))
```
taoyuanyuan, line 83: a request maybe call the func more than once

cch123: @taoyuanyuan , this is ok, the caller only return the pool when CheckAndInit returns true, and under this situation, the client slot must be inited and the index is saved to the ctx

cch123: currently, we'll set the client index to context when CheckAndInit returns true, and use this index to find the client in the pool should be valid

pkg/stream/xprotocol/connpool_multiplex.go
```diff
@@ -136,14 +159,17 @@ func (p *poolMultiplex) NewStream(ctx context.Context, receiver types.StreamRece
 
 // Shutdown stop the keepalive, so the connection will be idle after requests finished
 func (p *poolMultiplex) Shutdown() {
-	f := func(k, v interface{}) bool {
-		ac, _ := v.(*activeClientMultiplex)
-		if ac.keepAlive != nil {
-			ac.keepAlive.keepAlive.Stop()
+	atomic.StoreInt64(&p.shutdown, 1)
+	for i := 0; i < len(p.activeClients); i++ {
```
taoyuanyuan, line 172: for _, client := range p.activeclient

cch123: image

This grammar will cause the lock in sync.Map to be copied

pkg/stream/xprotocol/connpool_multiplex.go
```diff
@@ -36,47 +36,72 @@ import (
...
 	}, nil)
 }
 
 // CheckAndInit init the connection pool
 func (p *poolMultiplex) CheckAndInit(ctx context.Context) bool {
-	var client *activeClientMultiplex
+	var clientIdx int64 = 0 // most use cases, there will only be 1 connection
+	if len(p.activeClients) > 1 {
+		if clientIdx = getClientIDFromDownStreamCtx(ctx); clientIdx == invalidClientID {
+			clientIdx = atomic.AddInt64(&p.currentCheckAndInitIdx, 1) % int64(len(p.activeClients))
```
wangfakang, line 91: maybe better use `AddUint64`  ? because the value type of `Connections().Max()`  is `uint64`.

cch123: invalidClientID == -1 is used to check whether the client id is set in context

pkg/stream/xprotocol/connpool_multiplex.go
```diff
@@ -136,14 +174,26 @@ func (p *poolMultiplex) NewStream(ctx context.Context, receiver types.StreamRece
...
 // Shutdown stop the keepalive, so the connection will be idle after requests finished
 func (p *poolMultiplex) Shutdown() {
-	f := func(k, v interface{}) bool {
-		ac, _ := v.(*activeClientMultiplex)
-		if ac.keepAlive != nil {
-			ac.keepAlive.keepAlive.Stop()
+	utils.GoWithRecover(func() {
+		p.clientMux.Lock()
+		if p.shutdown {
+			return
```
nejisama, line 180: return without unlock?

cch123: good catch

pkg/stream/xprotocol/connpool_multiplex.go
```diff
@@ -87,19 +112,32 @@ func (p *poolMultiplex) CheckAndInit(ctx context.Context) bool {
...
+			// this client is not inited
+			return p.Host(), nil, types.ConnectionFailure
+		}
+	}
+
 	subProtocol := getSubProtocol(ctx)
 
-	client, _ := p.activeClients.Load(subProtocol)
-	host := p.Host()
+	client, _ := p.activeClients[clientIdx].Load(subProtocol)
```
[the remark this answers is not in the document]
cch123: only when CheckAndInit returns true, the getActiveConnectionPool returns this pool, so the clientIdx must be a valid value

taoyuanyuan (approved):
LGTM

Issue #1384, opened 2020-09-24
xDS received message larger than max


champly:
### Describe the bug

When the CRD too many, the xDS content will large than default 4M.

Our situation has 2100 CDS

```
2020-09-23 00:54:23,533 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5193322 vs. 4194304), retry after 1s
2020-09-23 00:54:24,533 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (405942322 vs. 4194304), retry after 1s
2020-09-23 00:54:25,534 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (758135085 vs. 4194304), retry after 1s
2020-09-23 00:54:26,534 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (861155376 vs. 4194304), retry after 1s
2020-09-23 00:54:27,534 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (892549680 vs. 4194304), retry after 1s
2020-09-23 00:54:28,534 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (1513043760 vs. 4194304), retry after 1s
2020-09-23 00:54:29,534 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (310510346 vs. 4194304), retry after 1s
2020-09-23 00:54:30,534 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (1954115685 vs. 4194304), retry after 1s
2020-09-23 00:54:31,535 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (1735356263 vs. 4194304), retry after 1s
2020-09-23 00:54:32,535 [INFO] [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (1700884585 vs. 4194304), retry after 1s
[...]

PR #1385, opened 2020-09-24
Fix xds resource exhausted

champly:

#1384 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


champly:
@wangfakang PTAL

pkg/xds/v2/xdsconfig.go
```diff
@@ -299,3 +299,14 @@ func (c *ADSConfig) closeADSStreamClient() {
...
 }
+
+// if got
+// [xds] [ads client] get resp timeout: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5193322 vs. 4194304), retry after 1s
+// this log, should change this size
+const xdsMaxRecvMsgSize = 4 * 1024 * 1024
+
+func generateDialOption() grpc.DialOption {
+	return grpc.WithDefaultCallOptions(
+		grpc.MaxCallRecvMsgSize(xdsMaxRecvMsgSize),
```
wangfakang, line 310: It would be better to add an xDS configuration item.

pkg/xds/v2/adssubscriber.go
```diff
@@ -84,7 +85,13 @@ func (adsClient *ADSClient) receiveThread() {
 			}
 			resp, err := sc.Recv()
 			if err != nil {
-				log.DefaultLogger.Infof("[xds] [ads client] get resp timeout: %v, retry after 1s", err)
+				// rpc error: code = Unavailable desc = transport is closing
+				if strings.Contains(err.Error(), "transport is closing") {
+					adsClient.reconnect()
```
wangfakang, line 90: maybe conflict with `adsClient.reconnect()` in `sendThread()`.

champly: remove it.

pkg/xds/v2/xdsconfig.go
champly: Set with gRPC default `math.MaxInt32`

wangfakang (approved):
LGTM.

Issue #1372, opened 2020-09-18
is this a bug smooth upgrade bug


alpha-baby:
### Describe the bug

https://github.com/mosn/mosn/blob/master/pkg/server/handler.go#L290

if `l.listener.bindToPort` value is `false`, so `l.listener.rawl` is nil.

And then smooth upgrade will be fail.

is this  a bug or feature or special design ？

### Minimal yet complete reproducer code (or GitHub URL to code)

```golang
func (ch *connHandler) ListListenersFile(lctx context.Context) []*os.File {
	files := make([]*os.File, len(ch.listeners))
	for idx, l := range ch.listeners {
		file, err := l.listener.ListenerFile()
		if err != nil {
			log.DefaultLogger.Alertf("listener.list", "[server] [conn handler] fail to get listener %s file descriptor: %v", l.listener.Name(), err)
			return nil //stop reconfigure
		}
		files[idx] = file
	}
	return files
}
```

### Environment

- MOSN Version
git master branch ， latest version

taoyuanyuan:
please upload your fail log.

alpha-baby:
> please upload your fail log.

@taoyuanyuan 

old mosn

```txt
2020-09-18 15:11:04,996 [ERROR] [listener.list] [server] [conn handler] fail to get listener 0.0.0.0_22000 file descriptor: invalid argument
2020-09-18 15:11:04,996 [INFO] [admin store] [mosn state] state changed to 1
```

new mosn

```txt
2020/09/18 15:11:04 load config from :  xds-config.json
2020-09-18 15:11:15,1 [ERROR] [server] InheritListeners Accept error :accept unix /Users/fujianhao3/Desktop/mediator/.mydevCachefile/listen.sock: i/o timeout
2020-09-18 15:11:15,2 [FATAL] [mosn] [NewMosn] getInheritListeners failed, exit
```

>fail to get listener 0.0.0.0_22000

this listenner is 

```goalng
type listener struct {
	name                    string
	localAddress            net.Addr
	bindToPort              bool 
	listenerTag             uint64
	perConnBufferLimitBytes uint32
	useOriginalDst          bool
	cb                      types.ListenerEventListener
	rawl                    *net.TCPListener
	config                  *v2.Listener
	mutex                   sync.Mutex
	// listener state indicates the listener's running state. The listener state effects if a listener binded to a port
	state ListenerState
}
```

`listener.bindToPort` value is`false`.

taoyuanyuan:
[...]

PR #1409, opened 2020-09-30
fix new bug for issue #1372

alpha-baby:

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/network/listener.go
+++ b/pkg/network/listener.go
@@ -265,13 +265,14 @@ func (l *listener) UseOriginalDst() bool {
 }
 
 func (l *listener) Close(lctx context.Context) error {
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	l.state = ListenerStopped
+
 	if !l.bindToPort {
 		return nil
 	}
 
-	l.mutex.Lock()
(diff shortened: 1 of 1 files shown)


pkg/network/listener.go
```diff
@@ -220,15 +233,28 @@ func (l *listener) SetListenerTag(tag uint64) {
 }
 
 func (l *listener) ListenerFile() (*os.File, error) {
+	err := errors.New("not support this network " + l.network)
```
nejisama, line 236: do not create an error every time. use bindport check instead?

nejisama:
```
if  ! l.bindport {
    return nil, syscall.EINVAL
}
```

nejisama (approved):

pkg/network/listener.go
```diff
@@ -256,6 +265,10 @@ func (l *listener) UseOriginalDst() bool {
 }
 
 func (l *listener) Close(lctx context.Context) error {
+	if !l.bindToPort {
```
nejisama, line 268: l.state = ListenerStopped should be setted too

PR #1427, opened 2020-10-21
Fix/tls hashvalue

nejisama:
### description
Fix TLS HashValue in host.

we use TLSHashValue to check wether the connection should be changed between tls and plaintext.
If a host config (with cluster config) is not support tls, the connection always plaintext.
If a host config (with cluster config) is support tls, the connection MAYBE tls, that based on the global variables.

A connpool cached ths TLSHashValue (with global variables) , but the host.CreateConnection is effected by global variables. so we should update connpool(create a new one) if host config (with cluster config) is changed even the connection is still plaintext (by the  global variables.)

image

### code

+ Core Code see `host.go`
+ UpdateClusterManagerTLS is a bug fix.
+ others are modified for write test case


wangfakang (approved):
LGTM.

PR #1463, opened 2020-11-05
Bug Fix: nil pointer ref

GLYASAI:
### Bug detail

The configuration of the mixer filter may look like this:
```
{
  "forward_attributes":
    {
      "attributes":
        {
          "source.namespace": { "string_value": "XYZ11" },
          "source.uid": { "string_value": "POD11" },
        },
    },
}
```

No Transport specified will result in a nil pointer reference.

In addition, when `h.serviceContext.serviceConfig` is nil, it will also cause nil pointer reference.

### Solutions

1. Create a default Transport for mixer filter.
2. Maker sure `serviceConfig` is not nil before `serviceConfig.DisableReportCalls`.

### DIFF
--- a/pkg/filter/stream/mixer/mixer.go
+++ b/pkg/filter/stream/mixer/mixer.go
@@ -173,5 +173,12 @@ func ParseMixerFilter(cfg map[string]interface{}) (*v2.Mixer, error) {
 		return nil, err
 	}
 
+	// default configuration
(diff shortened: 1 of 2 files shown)


taoyuanyuan (approved):
LGTM

champly (approved):
LGTM

PR #1496, opened 2020-11-24
pool the writebufferchan write timer for high load

cch123:


pkg/network/connection.go
```diff
@@ -575,10 +575,11 @@ func (c *connection) Write(buffers ...buffer.IoBuffer) (err error) {
...
 
 			// fail after 60s
-			t := time.NewTimer(types.DefaultConnTryTimeout)
+			t := acquireTimer(types.DefaultConnTryTimeout)
 			select {
 			case c.writeBufferChan <- &buffers:
 				t.Stop()
+				releaseTimer(t)
 			case <-t.C:
 				err = types.ErrWriteBufferChanTimeout
```
champly, line 584: should invoke `releasetTimer` put timer to `timerPool`

pkg/network/timer_pool.go
```diff
@@ -0,0 +1,30 @@
...
+	if t.Reset(d) {
+		return time.NewTimer(d)
+	}
+
+	return t
+}
+
+func releaseTimer(t *time.Timer) {
+	if !t.Stop() {
+		<-t.C
```
nejisama, line 27:
```
Note that it is not possible to use Reset's return value correctly, as there is a race condition between draining the channel and the new timer expiring.
```
Try this 
```
if !t.Stop() && len(t.C) > 0 {
   <-t.C
}
```

cch123:
move to this PR
https://github.com/mosn/mosn/pull/1498

PR #1506, opened 2020-12-03
Auto protocol match support xprotocol

cadeeper:
auto match support xprotocol

### UT result
截屏2020-12-03 下午4 21 32

### DIFF
--- a/pkg/protocol/xprotocol/factory.go
+++ b/pkg/protocol/xprotocol/factory.go
@@ -66,17 +66,8 @@ func GetMatcher(name types.ProtocolName) types.ProtocolMatch {
 }
 
 //GetMatchers return all matchers that was registered
-func GetMatchers() []types.ProtocolMatch {
-	if len(matcherMap) == 0 {
-		return nil
-	}
-	var matchers = make([]types.ProtocolMatch, len(matcherMap))
-	i := 0
-	for _, matcher := range matcherMap {
-		matchers[i] = matcher
-		i++
-	}
-	return matchers
+func GetMatchers() map[types.ProtocolName]types.ProtocolMatch {
+	return matcherMap
 }
 
 // RegisterMapping register the HTTP status code mapping function of the protocol into factory
--- a/pkg/protocol/xprotocol/factory.go
+++ b/pkg/protocol/xprotocol/factory.go
@@ -65,6 +65,20 @@ func GetMatcher(name types.ProtocolName) types.ProtocolMatch {
 	return matcherMap[name]
 }
 
+//GetMatchers return all matchers that was registered
+func GetMatchers() []types.ProtocolMatch {
+	if len(matcherMap) == 0 {
+		return nil
+	}
+	var matchers = make([]types.ProtocolMatch, len(matcherMap))
+	i := 0
+	for _, matcher := range matcherMap {
+		matchers[i] = matcher
+		i++
+	}
+	return matchers
+}
+
 // RegisterMapping register the HTTP status code mapping function of the protocol into factory
 func RegisterMapping(name types.ProtocolName, mapping protocol.HTTPMapping) error {
 	// check name conflict
--- a/pkg/stream/xprotocol/factory.go
+++ b/pkg/stream/xprotocol/factory.go
@@ -54,14 +54,18 @@ func (f *streamConnFactory) ProtocolMatch(context context.Context, prot string, 
 	if subProtocolMatchers == nil {
 		return stream.FAILED
 	}
+	again := false
 	for _, matcher := range subProtocolMatchers {
 		result := matcher(magic)
 		if result == types.MatchSuccess {
 			return nil
 		}
 		if result == types.MatchAgain {
-			return stream.EAGAIN
+			again = true
 		}
+	}
+	if again {
+		return stream.EAGAIN
 	}
 	return stream.FAILED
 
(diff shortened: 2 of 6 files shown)


pkg/stream/xprotocol/factory.go
```diff
@@ -49,5 +50,23 @@ func (f *streamConnFactory) CreateBiDirectStream(context context.Context, connec
...
+		result := matcher(magic)
+		if result == types.MatchSuccess {
+			return nil
+		}
+		if result == types.MatchAgain {
+			again = true
+		}
+	}
+	if again {
+		return stream.EAGAIN
```
champly, line 68: If the protocol is unknown, always return `stream.EAGAIN` will fall into dead loop.

pkg/protocol/xprotocol/factory.go
```diff
@@ -65,6 +65,20 @@ func GetMatcher(name types.ProtocolName) types.ProtocolMatch {
 	return matcherMap[name]
 }
 
+//GetMatchers return all matchers that was registered
+func GetMatchers() []types.ProtocolMatch {
```
champly, line 69: Should build `matchers` with `RegisterMatcher`, each request build `matchers` will cost performance.

cadeeper: Matchers may be changed by the caller, for this reason, I build it every request. I'll optimize it.

cadeeper: > Should build `matchers` with `RegisterMatcher`, each request build `matchers` will cost performance.

Matchers may be changed by the caller, for this reason, I build it every request. I'll optimize it.

champly: Why not return `matcherMap` direct?

cadeeper: > Why not return `matcherMap` direct?

To protect matcherMap from change by the caller

pkg/stream/xprotocol/factory.go
cadeeper: > If the protocol is unknown, always return `stream.EAGAIN` will fall into dead loop.

I don't understand. stream.EAGAIN only happened that subProtocolMatchers returns types.MatchAgain. if the protocol is unknown means each subProtocolMatchers will finally return types.MatchFailed. Then ProtocolMatch will return stream.FAILED

pkg/protocol/xprotocol/factory.go
cadeeper: > Why not return `matcherMap` direct?

If it's not necessary.I'll return matcherMap direct.

pkg/stream/xprotocol/factory.go
```diff
@@ -49,5 +50,23 @@ func (f *streamConnFactory) CreateBiDirectStream(context context.Context, connec
 }
 
 func (f *streamConnFactory) ProtocolMatch(context context.Context, prot string, magic []byte) error {
+	subProtocolMatchers := xprotocol.GetMatchers()
```
taoyuanyuan, line 53: don't need configure `sub_protocol` by set "types.ContextSubProtocol"

cadeeper: done

cadeeper: > don't need configure `sub_protocol` by set "types.ContextSubProtocol"

done

taoyuanyuan:
change base branch to master

taoyuanyuan (dismissed):
LGTM

champly (approved):
LGTM

taoyuanyuan:
it's wrong when configure upstreamProtocol auto.
https://github.com/mosn/mosn/blob/916254bc42db353adf9ea792347180de1d0d0f3b/pkg/stream/xprotocol/connpool.go#L100
https://github.com/mosn/mosn/blob/916254bc42db353adf9ea792347180de1d0d0f3b/pkg/stream/xprotocol/connpool_binding.go#L231

taoyuanyuan:
we could delete `func getProtocol()`, and set `protocol.Xprotocol` directly

cadeeper:
> we could delete `func getProtocol()`, and set `protocol.Xprotocol` directly

What will happen if it is not set to protocol.Xprotocol directly? this will set to the same as downstream's protocol if upstream protocol is AUTO:
https://github.com/mosn/mosn/blob/916254bc42db353adf9ea792347180de1d0d0f3b/pkg/proxy/downstream.go#L716

taoyuanyuan:
> > we could delete `func getProtocol()`, and set `protocol.Xprotocol` directly
> 
> What will happen if we don't set to protocol.Xprotocol directly? this will set to the same as downstream's protocol if upstream protocol is AUTO:
> https://github.com/mosn/mosn/blob/916254bc42db353adf9ea792347180de1d0d0f3b/pkg/proxy/downstream.go#L716

The result is the same， the func getProtocol return `protocol.Xprotocol` certainly.

cadeeper:
> > > we could delete `func getProtocol()`, and set `protocol.Xprotocol` directly
> > 
> > 
> > What will happen if we don't set to protocol.Xprotocol directly? this will set to the same as downstream's protocol if upstream protocol is AUTO:
> > https://github.com/mosn/mosn/blob/916254bc42db353adf9ea792347180de1d0d0f3b/pkg/proxy/downstream.go#L716
> 
> The result is the same， the func getProtocol return `protocol.Xprotocol` certainly.

done

taoyuanyuan (approved):
LGTM

taoyuanyuan:
thx

PR #1509, opened 2020-12-07
No provider available occurred after dubbo2.7.3

cadeeper:
As the title mentioned. After dubbo2.7.3, 'default.' prefix was removed from URL key. Mosn always pass version and group parameters with empty string by default. But unfortunately consumer set null by default. they won't match when consumer do subscribe. 
See: org.apache.dubbo.common.utils.UrlUtils#isMatch

### Solutions
Don't pass version and group parameter by default when provider register.


cadeeper:
PTAL

champly (approved):
LGTM

cch123 (approved):
LGTM

nejisama:
DO NOT merge code into `develop` branch if you are not sure why this code should merged into `develop`. 
create pull request to master. this merge will be revert. @cadeeper

PR #1512, opened 2020-12-08
fix(conn): change the read err log from info to debug

Sharember:

KeepAlive connections without data transfer will frequently print timeout logs and eventually fill the disk， like:

image

so, need change the read err log from info to debug.

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


nejisama:
use #1513 instead. the branch `develop` planned to be merged into master this week

nejisama:
this log is delete by #1513, 
and merged into master by #1527 
close this pr.
thanks

Issue #1518, opened 2020-12-13
not exists if a key with an empty value in api.HeaderMap


GLYASAI:
### Describe the bug

A key with an empty value set to api.HeaderMap. But header.Get(key) thinks that the key does not exist. 

```
header := http.RequestHeader{
	RequestHeader: &fasthttp.RequestHeader{},
}
header.Set("foobar", "")
_, exists := header.Get("foobar")
```
 Expected exists to be true, but got false.

ianwoolf:
the method of `mosn/pkg/protocol/http`.`RequestHeader` has no pointer reciever, and its method `Set` has a problem.  i'm working on it

nejisama:
fixed

wangfakang:
TODO 
upgrade fasthttp:
https://github.com/valyala/fasthttp/commit/4e63057c0fd0ebcc9040c161ad69c7e56d57cf82

PR #1532, opened 2020-12-24
fix invalid empty value cause by nil EmptyValueHeaders

ianwoolf:

#1518 

### Solutions

function `Set` has no pointer reciever, so  `make EmptyValueHeaders` in `Set` does not take effect. So, add a New function to create struct which with a alread maked map to fix it.

### UT result

add test to pkg/protocol/http/types_test.go

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/protocol/http/types.go
+++ b/pkg/protocol/http/types.go
@@ -85,41 +85,40 @@ const (
 	LoopDetected                  = 508
 	NotExtended                   = 510
 	NetworkAuthenticationRequired = 511
+
+	PlaceHolder = "-"
 )
 
 type RequestHeader struct {
 	*fasthttp.RequestHeader
-
-	// Due to the fact that fasthttp's implementation doesn't have correct semantic for Set("key", "") and Peek("key") at the
-	// first time of usage. We need another way for compensate.
-	//
-	// The problem is caused by the func initHeaderKV, if the original kv.value is nil, ant input value is also nil,
-	// then the final kv.value remains nil.
-	//
-	// kv.value = append(kv.value[:0], value...)
-	//
-	// fasthttp do has the kv entry, but kv.value is nil, so Peek("key") return nil. But we want "" instead.
-	EmptyValueHeaders map[string]bool
 }
 
 // Get value of key
 func (h RequestHeader) Get(key string) (string, bool) {
 	result := h.Peek(key)
-	if result != nil || h.EmptyValueHeaders[key] {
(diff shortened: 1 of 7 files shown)


ianwoolf:
the former emtpyValue test is pass because `clone()` make the `EmptyValueHeaders`. So i create a new test func

pkg/protocol/http/types.go
```diff
@@ -85,41 +85,29 @@ const (
...
 
 // Set key-value pair in header map, the previous pair will be replaced if exists
 func (h RequestHeader) Set(key string, value string) {
-	h.RequestHeader.Set(key, value)
 	if value == "" {
-		if h.EmptyValueHeaders == nil {
-			h.EmptyValueHeaders = make(map[string]bool)
-		}
-		h.EmptyValueHeaders[key] = true
+		h.RequestHeader.Set(key, PlaceHolder)
```
taoyuanyuan, line 108: add comment

pkg/protocol/http/types.go
```diff
@@ -85,41 +85,40 @@ const (
...
+//
+// fasthttp do has the kv entry, but kv.value is nil, so Peek("key") return nil. But we want "" instead.
 func (h RequestHeader) Set(key string, value string) {
-	h.RequestHeader.Set(key, value)
 	if value == "" {
-		if h.EmptyValueHeaders == nil {
-			h.EmptyValueHeaders = make(map[string]bool)
-		}
-		h.EmptyValueHeaders[key] = true
+		// Set a placeholder first, so that ResponseHeader can get this value after setting an emtpy value.
```
taoyuanyuan, line 118: ResponseHeader  - > RequestHeader

pkg/protocol/http/types.go
taoyuanyuan: ResponseHeader ->    RequestHeader

ianwoolf: 👍      i fix it

ianwoolf (commented):
👍     i fix the comment

pkg/protocol/http/types.go
ianwoolf: 👍 i fix the comment

ianwoolf: 👍 i fix the comment

taoyuanyuan (approved):
LGTM

nejisama (approved):

PR #1570, opened 2021-02-01
fix: fix unstable test cases for flow control filter

ansiz:

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


wangfakang (approved):
LGTM.

antJack (dismissed):
LGTM

pkg/filter/stream/flowcontrol/stream_filter_test.go
```diff
@@ -70,8 +70,15 @@ func TestStreamFilter(t *testing.T) {
 			}, nil, 0))
```
wangfakang, line 70: need fix gethandler for the variable.
```
        variable.RegisterVariable(variable.NewBasicVariable(k, nil,
            func(ctx context.Context, variableValue *variable.IndexedValue, data interface{}) (s string, err error) {
                val := m[k]
                return val, nil
            }, nil, 0))
```

wangfakang (approved):

taoyuanyuan (approved):
LGTM

PR #1576, opened 2021-02-04
Feature add MOSN network extentions.

wangfakang:
FYI https://github.com/mosn/mosn/issues/1563

# Quic start

#### Install mosn network on envoy image

```
docker pull mosnio/mosn-network-on-envoy:v0.20.0
```

#### Start mosn network on envoy container

```
docker run -it --rm --net=host mosnio/mosn-network-on-envoy:v0.20.0
```

#### Test dynamic route sample

In this sample, after set Envoy as a network extension for MOSN , then Envoy's dynamic routing feature can be implemented by developing a stream filter on the MOSN side using the GoLang language.

For example, requests with `uid` range `[1,100]` will be routed to `test` app site `s1`, requests with `uid` range `[101, 200]` will be routed to site `s2`, requests with other `uid` ranges will be routed to site `s3`, and if there is no `uid` header, the request will be round robin to `s1`, `s2`, and `s3`.

Then we could see the route rules at work by making a request via `curl` in another terminal:

```
$curl localhost:2990 -H "uid:2"
site s1 from 3450

$curl localhost:2990 -H "uid:2"
site s1 from 3451

$curl localhost:2990 -H "uid:200"
site s2 from 3452

$curl localhost:2990 -H "uid:201"
site s3 from 3453
```


taoyuanyuan (approved):
LGTM

antJack (approved):
LGTM

PR #1596, opened 2021-02-25
fix: duplicate called filepath.clean method in streamfilter config

eliasyaoyc:
### Solutions
duplicate called filepath.clean method in streamfilter config.
because clean method will be called in filepath.abs, so don't called again.
image


antJack (approved):
LGTM

taoyuanyuan (approved):
LGTM

PR #1641, opened 2021-04-27
[fix] fix up mem leak problem for sm4 decode error

william-zk:
1. fix up mem leak problem for sm4 decode error
2. rewrite code with a better code style


taoyuanyuan (approved):
LGTM

nejisama (approved):

PR #1741, opened 2021-08-07
feat: wasm manager support reload

zu1k:

https://github.com/mosn/layotto/issues/165

### Solutions

Add func `ReloadWasmByName` for wasm manager to provide the ability to reload a wasm plugin without modify the config.

This func will be used by `layotto` to achieve wasm hot reload for developer.

### UT result

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/config/v2/wasm.go
+++ b/pkg/config/v2/wasm.go
@@ -27,6 +27,7 @@ type WasmVmConfig struct {
 	Engine string `json:"engine,omitempty"`
 	Path   string `json:"path,omitempty"`
 	Url    string `json:"url,omitempty"`
+	Md5    string `json:"md5,omitempty"`
 	Cpu    int    `json:"cpu,omitempty"`
 	Mem    int    `json:"mem,omitempty"`
 }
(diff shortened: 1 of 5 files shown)


seeflood:
@ nejisama @taoyuanyuan Hi guys, could you please review this pr?
Currently Layotto needs this feature for hot reloading wasm modules  :>

antJack:
Is it possible to implement this feature by simply checking the MD5 or some other hash value of the .wasm file inside WasmManager.AddOrUpdatePlugin?
① checking of hash value could enhance the security of the whole wasm framework, as well as prevent file corruption caused by concurrent changing of the .wasm file
② I feel a little confused about when to call the ForceUpdate/AddOrUpdatePlugin by just seeing those APIs.

zu1k:
I have re-implemented this by using md5.

Where the plugin already exists, it will first check the md5 of the new configuration, force reload if not specified, and update the md5 value in the configuration.

If the md5 is not empty in the new configuration, compare it with the old value to determine if it needs to be reloaded. The real md5 value of the file is calculated on reload to ensure that the loaded file is correct.

antJack (approved):
LGTM

taoyuanyuan (approved):
LGTM

taoyuanyuan:
thx ~

PR #1769, opened 2021-09-06
fix boltv2 heartbeat and trace

nejisama:
1. Fix boltv2 heartbeat trigger and reply with boltv2 request / response.
2. Fix boltv2 hijack consturct
3. Add boltv2 trace delegate: Set SPAN_TYPE in delegate, or the trace will be ingored.

### DIFF
--- a/pkg/protocol/xprotocol/boltv2/protocol.go
+++ b/pkg/protocol/xprotocol/boltv2/protocol.go
@@ -130,7 +130,7 @@ func (proto *boltv2Protocol) Decode(ctx context.Context, data types.IoBuffer) (i
 func (proto *boltv2Protocol) Trigger(ctx context.Context, requestId uint64) api.XFrame {
 	return &Request{
 		RequestHeader: RequestHeader{
-			Version1: 0x01,
+			Version1: ProtocolVersion1,
 			RequestHeader: bolt.RequestHeader{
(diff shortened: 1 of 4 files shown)


pkg/protocol/xprotocol/boltv2/protocol.go
```diff
@@ -125,22 +125,47 @@ func (proto *boltv2Protocol) Decode(ctx context.Context, data types.IoBuffer) (i
...
 // heartbeater
-// boltv2 send bolt heartbeat
+// set version1 as default 0x01, no crc, use version 1
+// see details in: https://github.com/sofastack/sofa-bolt/blob/master/src/main/java/com/alipay/remoting/rpc/protocol/RpcCommandEncoderV2.java
 func (proto *boltv2Protocol) Trigger(ctx context.Context, requestId uint64) api.XFrame {
-	engine := xprotocol.GetProtocol(bolt.ProtocolName)
-	return engine.Trigger(ctx, requestId)
+	return &Request{
+		RequestHeader: RequestHeader{
+			Version1: 0x01,
```
wangfakang, line 133: `0x01` appears in multiple places, so it might be better to use a constant variable.

wangfakang (approved):

taoyuanyuan (approved):
lgtm

PR #1793, opened 2021-09-27
upgrade proxy-wasm-go-host

zhenjunMa:

Mosn previously relied on proxy-wasm-go-host v0.1.0, which is quite different from the latest version of proxy-wasm-go-host, but mosn has not been upgraded. In the future, the expansion of wasm will be based on the new version of proxy-wasm-go -host, so we need to upgrade proxy-wasm-go-host to the latest version.

### Solutions

1. Upgrade the version of proxy-wasm-go-host used
2. Adjust the code to adapt to the new version of proxy-wasm-go-host

### Code Style
+ Show `Golint` result


antJack (approved):
LGTM

taoyuanyuan (approved):
lgtm

PR #1797, opened 2021-10-04
add IPAllowlist filter

Bryce-huang:
add IPAllowlist filter

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


taoyuanyuan:
> After referring to nginx, I think that blocklist and allowlist can be used at the same time. When both are configured, blocklist will be used to check whether ip is forbidden, and then allowlist is used to check whether ip is allowed to be accessed. When only one of them is configured, only the configured one will take effect. Do you think this satisfies the demand?

I think that blocklist and allowlist can be configured more than one，and the rules are checked in sequence until the first match is found.

taoyuanyuan:
> > After referring to nginx, I think that blocklist and allowlist can be used at the same time. When both are configured, blocklist will be used to check whether ip is forbidden, and then allowlist is used to check whether ip is allowed to be accessed. When only one of them is configured, only the configured one will take effect. Do you think this satisfies the demand?
> 
> I think that blocklist and allowlist can be configured more than one，and the rules are checked in sequence until the first match is found.

oh, you are right！our list can be confirugred more than one addrs.
btw,  what‘s the default action when both are configured, allow or deny？

Bryce-huang:
> > After referring to nginx, I think that blocklist and allowlist can be used at the same time. When both are configured, blocklist will be used to check whether ip is forbidden, and then allowlist is used to check whether ip is allowed to be accessed. When only one of them is configured, only the configured one will take effect. Do you think this satisfies the demand?
> 
> I think that blocklist and allowlist can be configured more than one，and the rules are checked in sequence until the first match is found.



> > > After referring to nginx, I think that blocklist and allowlist can be used at the same time. When both are configured, blocklist will be used to check whether ip is forbidden, and then allowlist is used to check whether ip is allowed to be accessed. When only one of them is configured, only the configured one will take effect. Do you think this satisfies the demand?
> > 
> > 
> > I think that blocklist and allowlist can be configured more than one，and the rules are checked in sequence until the first match is found.
> 
> oh, you are right！our list can be confirugred more than one addrs. But, what‘s the default action that both are configured, allow or deny？

deny, blocklist first

taoyuanyuan:
> > > After referring to nginx, I think that blocklist and allowlist can be used at the same time. When both are configured, blocklist will be used to check whether ip is forbidden, and then allowlist is used to check whether ip is allowed to be accessed. When only one of them is configured, only the configured one will take effect. Do you think this satisfies the demand?
> > 
> > 
> > I think that blocklist and allowlist can be configured more than one，and the rules are checked in sequence until the first match is found.
> 
> > > > After referring to nginx, I think that blocklist and allowlist can be used at the same time. When both are configured, blocklist will be used to check whether ip is forbidden, and then allowlist is used to check whether ip is allowed to be accessed. When only one of them is configured, only the configured one will take effect. Do you think this satisfies the demand?
> > > 
> > > 
> > > I think that blocklist and allowlist can be configured more than one，and the rules are checked in sequence until the first match is found.
> > 
> > 
> > oh, you are right！our list can be confirugred more than one addrs. But, what‘s the default action that both are configured, allow or deny？
> 
> deny, blocklist first

if we have a scenario like this，`allow` will be default action. it's too flexible...
```
    allow  192.168.1.1;
    deny 192.168.1.0/24;
    allow all;
```

taoyuanyuan:
Welcome to our community （dingding）
image

PR #1834, opened 2021-11-09
support admin api extends auth functions

nejisama:
1. Support a new way to registered enhanced admin api functions. keep compatbile for original ways.
2. Export the admin api functions

### DIFF
--- a/pkg/admin/server/handler.go
+++ b/pkg/admin/server/handler.go
@@ -43,9 +43,12 @@ func NewAuth(check func(*http.Request) bool, failed func(http.ResponseWriter)) *
 }
 
 func (a Auth) Check(w http.ResponseWriter, r *http.Request) bool {
-	pass := a.checkAction(r)
-	if !pass {
-		a.failedAction(w)
+	pass := true
+	if a.checkAction != nil {
+		pass = a.checkAction(r)
+		if !pass {
+			a.failedAction(w)
+		}
 	}
 	return pass
 }
--- a/pkg/admin/server/handler.go
+++ b/pkg/admin/server/handler.go
@@ -23,15 +23,40 @@ import (
 	"mosn.io/mosn/pkg/log"
 )
 
+// Auth contains two parts: check function and failed function.
+// if check function returns false, the Auth.Check will returns false and
+// call failed function. if failed function is nil, use default instead.
+// default function will write a http forbidden without body
+type Auth struct {
+	checkAction  func(*http.Request) bool
+	failedAction func(http.ResponseWriter)
+}
+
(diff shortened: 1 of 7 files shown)


pkg/admin/server/handler.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"mosn.io/mosn/pkg/log"
+)
+
+// APIHandler is a wrapper of http.Handler, which contains auth options
+type APIHandler struct {
+	// the real handler function
+	handler func(http.ResponseWriter, *http.Request)
+	// chains for auth, one of the auth returns false, the request will be failed.
+	auths []func(*http.Request) bool
```
taoyuanyuan, line 31: One handler belongs to one user， so I think that no need to use chains.

taoyuanyuan: maybe `auths` and `filead` could combine into one function.
`auths func(w http.ResponseWriter, r *http.Request) bool`?

nejisama: auths functions can be minimized. for example, different auth functions can be combined for different handler

pkg/admin/server/handler.go
```diff
@@ -23,25 +23,49 @@ import (
...
 type APIHandler struct {
 	// the real handler function
 	handler func(http.ResponseWriter, *http.Request)
-	// chains for auth, one of the auth returns false, the request will be failed.
-	auths []func(*http.Request) bool
-	// write repsonse when the request failed by auth function
-	// default is 403 Forbidden without body
-	failed func(http.ResponseWriter)
+	// chains for auth action, different auth can be combined for different handler.
+	// one of the auths check returns false, the
```
taoyuanyuan, line 58: Incomplete comment.

nejisama: fixed

pkg/admin/server/handler.go
```diff
@@ -0,0 +1,84 @@
...
+	checkAction  func(*http.Request) bool
+	failedAction func(http.ResponseWriter)
+}
+
+func NewAuth(check func(*http.Request) bool, failed func(http.ResponseWriter)) *Auth {
+	if failed == nil {
+		failed = defaultFailedFunc
+	}
+	return &Auth{
+		checkAction:  check,
```
taoyuanyuan, line 40: judge if `check` is nil

nejisama: check is required，i think tif check is nil causes panic is acceptable.

taoyuanyuan: should panic at start up，not at running.

taoyuanyuan (approved):
LGTM

wangfakang (approved):

PR #1839, opened 2021-11-10
fix issue #1833

MengJiapeng:
#1833

### Solutions
when find listener with originalDst IP:
1. use the listener that best match when found multi listeners;
2. close the connection if can't find any matching listener.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -642,7 +642,6 @@ func (arc *activeRawConn) SetOriginalAddr(ip string, port int) {
 
 func (arc *activeRawConn) UseOriginalDst(ctx context.Context) {
 	var listener, localListener *activeListener
-	var found bool
 
 	for _, lst := range arc.activeListener.handler.listeners {
 		if lst.listenIP == arc.originalDstIP && lst.listenPort == arc.originalDstPort {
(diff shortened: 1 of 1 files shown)


pkg/server/handler.go
```diff
@@ -666,28 +665,26 @@ func (arc *activeRawConn) UseOriginalDst(ctx context.Context) {
...
-	if !found {
-		if log.DefaultLogger.GetLogLevel() >= log.INFO {
-			log.DefaultLogger.Infof("[server] [conn] original dst:%s:%d", arc.activeListener.listenIP, arc.activeListener.listenPort)
-		}
-		arc.activeListener.OnAccept(arc.rawc, false, arc.oriRemoteAddr, ch, buf)
+	if log.DefaultLogger.GetLogLevel() >= log.ERROR {
+		log.DefaultLogger.Errorf("[server] [conn] can't found any listener by original dst: %s:%d", arc.originalDstIP, arc.originalDstPort)
 	}
+	// if we can't find any matching listener, close the connection
+	arc.rawc.Close()
```
taoyuanyuan, line 687: maybe we should keep the original logic and use the self listener.

taoyuanyuan (approved):
LGTM

antJack (approved):
LGTM

taoyuanyuan:
thank you for your contribution！

PR #1856, opened 2021-11-26
Admin API support Auth Extension and HTTP default header size set to 8K

nejisama:
1. ADMIN API support Auth Extension.
2. HTTP default max header size set to 8K


taoyuanyuan (approved):
LGTM

antJack (approved):
LGTM

PR #1893, opened 2021-12-23
bugfix: got the wrong CACert filename when converting the listen filter from istio, mosn may not listen success.

doujiang24:

got the wrong CACert filename when converting the listen filter from istio, mosn may not listen success.

### Solutions
used Filename instead of String.


taoyuanyuan (approved):
LGTM

wangfakang (approved):
LGTM

PR #1894, opened 2021-12-23
fix STRICT_DNS_CLUSTER ResolveTarget cannot stop

bincherry:
### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/upstream/cluster/strict_dns_cluster.go
+++ b/pkg/upstream/cluster/strict_dns_cluster.go
@@ -249,16 +249,17 @@ func (rt *ResolveTarget) StartResolve() {
 	for {
 		select {
 		case <-rt.stop:
-			rt.resolveTimeout.Stop()
-			rt.resolveTimer.Stop()
+			if log.DefaultLogger.GetLogLevel() >= log.DEBUG {
+				log.DefaultLogger.Debugf("[upstream] [strict dns cluster] stop resolve dns timer, address:%s, ttl:%d", rt.dnsAddress)
+			}
+			return
 		default:
 			select {
 			case <-rt.stop:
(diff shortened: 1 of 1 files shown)


pkg/upstream/cluster/strict_dns_cluster.go
```diff
@@ -251,6 +251,7 @@ func (rt *ResolveTarget) StartResolve() {
 		case <-rt.stop:
 			rt.resolveTimeout.Stop()
 			rt.resolveTimer.Stop()
+			return
```
wangfakang, line 254: Thank you for fixing it.
The following [case <-rt.stop](https://github.com/mosn/mosn/pull/1894/files#diff-c36f7a5b9260c86ee2d12759eb6b391caf06b96ed271f4fe52b285ce32ed90e1R257) can also be returned directly, thus reducing a select.

pkg/upstream/cluster/strict_dns_cluster.go
```diff
@@ -251,6 +251,7 @@ func (rt *ResolveTarget) StartResolve() {
 		case <-rt.stop:
 			rt.resolveTimeout.Stop()
 			rt.resolveTimer.Stop()
```
doujiang24, line 253: better to remove these two lines? since they are already in the defer function.

doujiang24: Also, better to add the debug log here? just like the next one.

doujiang24 (commented):
Nice Catch.

doujiang24 (approved):

wangfakang (approved):
LGTM

bincherry:
Thanks for your help.

PR #1905, opened 2021-12-31
chore: used the "/var/run/" path instead of "/dev/shm/" path for supervisor in docker image.

doujiang24:

May cause conflict in different containers,
since /dev/shm is the shared volume between different containers in the single pod.

to make the basic image works in this case:
https://mosn.io/blog/posts/mosn-sidecarset-hotupgrade/


nejisama (approved):
LGTM

taoyuanyuan (approved):

Issue #1906, opened 2021-12-31
unexpected handler of sigal interrupt


cuiweixie:
### Describe the bug

if catchSignalsCrossPlatform handle SIGINT with a little long time, handler in catchSignalsPosix will exit quickly without waitting catchSignalsCrossPlatform to exit.

### Environment

- MOSN Version

nejisama:
The signal-related logic is being reconstructed, and we no longer directly provide signal-related processing
see https://github.com/mosn/mosn/pull/1859

doujiang24:
@cuiweixie Already fixed in https://github.com/mosn/mosn/pull/1859. Thanks, closing it.

PR #1907, opened 2021-12-31
bugfix: unexpected handler of signal interrupt

cuiweixie:
#1906 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/server/keeper/serverkeeper.go
+++ b/pkg/server/keeper/serverkeeper.go
@@ -109,7 +109,14 @@ func catchSignalsCrossPlatform() {
 			case syscall.SIGHUP:
 				executeSignalCallback(syscall.SIGHUP)
 			case syscall.SIGINT:
+				// stop to quit
+				exitCode := ExecuteShutdownCallbacks("SIGINT")
+				for _, f := range onProcessExit {
+					f() // only perform important cleanup actions
+				}
+				//Stop()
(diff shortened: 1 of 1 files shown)


doujiang24:
Nice catch! I think we don't need to catch the `INT` signal in both  `catchSignalsPosix` and `catchSignalsCrossPlatform` since `syscall.SIGINT` eq `os.Interrupt`.

I have fixed it in another big PR in: https://github.com/mosn/mosn/pull/1859/files#diff-ff855c91e9da873e07593105f8123e6b43204398fc199f1d141481c85b8b57cfR73
It will be very helpful if you have time to help review it.

cuiweixie:
press ctrl+c twice the catchSignalsPosix will cause exit forcely without waiting catchSignalsCrossPlatform. at the beginning, I think i could be some posix spec. but it seems not after I google it just now.

doujiang24:
> press ctrl+c twice the catchSignalsPosix will cause exit forcely without waiting catchSignalsCrossPlatform. at the beginning, I think i could be some posix spec. but it seems not after I google it just now.

I think it's just some friendly usage, not necessary. I prefer to make it simpler.

doujiang24:
@cuiweixie Already fixed in https://github.com/mosn/mosn/pull/1859. Thanks, closing it.

PR #1917, opened 2022-01-07
[imp]ignore target dir of dubbo example code

scaat:

### Solutions
本地跑dubbo demo时，会生成对应的target目录，应该将其忽略掉。

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .*
 !./.travis.yml
-!./.gitignore
 build/bundles/
 *.log
 IMAGEBUILD/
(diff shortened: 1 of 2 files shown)


doujiang24 (commented):
maybe it's better to add a `.gitignore` under `examples/codes/dubbo`?

scaat:
OK

.gitignore
```diff
@@ -1,6 +1,5 @@
 .*
 !./.travis.yml
-!./.gitignore
```
doujiang24, line 3: I don't think we need to modify this line, you can use `git add --force` to add the newly added `.gitignore` file.
Or, we can just remove this line.

doujiang24 (approved):

PR #1933, opened 2022-01-17
opt filtermanager 'onContinueReading' performance

alpha-baby:

#1932 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -53,7 +53,7 @@ require (
 	google.golang.org/genproto v0.0.0-20200806141610-86f49bd18e98
 	google.golang.org/grpc v1.38.0
 	google.golang.org/grpc/examples v0.0.0-20210818220435-8ab16ef276a3
-	google.golang.org/protobuf v1.25.0
+	google.golang.org/protobuf v1.25.0 // indirect
 	istio.io/api v0.0.0-20200227213531-891bf31f3c32
 	k8s.io/klog v1.0.0
 	mosn.io/api v0.0.0-20211217011300-b851d129be01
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -64,3 +64,5 @@ require (
 )
 
 replace github.com/envoyproxy/go-control-plane => github.com/envoyproxy/go-control-plane v0.9.4
+
+replace mosn.io/api => ../mosn-api
```
doujiang24, line 68: should not push these local changes.

pkg/network/filtermanager.go
```diff
@@ -69,48 +69,43 @@ func (fm *filterManager) InitializeReadFilters() bool {
...
-				return
+				return false
 			}
 		}
+	}
+	return true
+}
 
-		buf := fm.conn.GetReadBuffer()
+func (fm *filterManager) onContinueReading(previousIndex int, buf buffer.IoBuffer) {
```
doujiang24, line 90:
```suggestion
func (fm *filterManager) onContinueReading(index int, buf buffer.IoBuffer) {
```
maybe we can just use `index` here? It looks more simpiler.

pkg/network/filtermanager.go
```diff
@@ -69,48 +69,43 @@ func (fm *filterManager) InitializeReadFilters() bool {
...
-			status := uf.filter.OnData(buf)
+	if previousIndex != 0 {
+		index = previousIndex + 1
+	}
 
-			if status == api.Stop {
-				//fm.conn.Write("your data")
-				return
-			}
+	for ; index < len(fm.upstreamFilters); index++ {
```
doujiang24, line 98: I'm not sure if it is safe to skip checking `uf.initialized` here.
@taoyuanyuan @nejisama may confirm it.

PR #1946, opened 2022-01-21
feature: support manipulate headers with variables

MengJiapeng:

### Solutions
通过路由规则新增/修改头部时，支持使用变量动态设置值；变量命名规则同accesslog的format表达式

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/router/headerformatter.go
+++ b/pkg/router/headerformatter.go
@@ -18,22 +18,22 @@
 package router
 
 import (
-	"mosn.io/api"
+	"context"
 	"mosn.io/mosn/pkg/variable"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"mosn.io/mosn/pkg/log"
-	"mosn.io/mosn/pkg/types"
 )
 
-var variableValueRegexp = regexp.MustCompile(`%.+%`)
+var variableValueRegexp = regexp.MustCompile(`^%.+%$`)
 
 func getHeaderFormatter(value string, append bool) headerFormatter {
 	if variableValueRegexp.MatchString(value) {
 		variableName := strings.Trim(value, "%")
-		if _, err := variable.Check(variableName); err != nil {
+		// todo cache the variable so we don't need to find it in format method
+		_, err := variable.Check(variableName)
+		if err != nil {
 			if log.DefaultLogger.GetLogLevel() >= log.WARN {
 				log.DefaultLogger.Warnf("invalid variable header: %s", value)
 			}
@@ -66,7 +66,7 @@ func (f *plainHeaderFormatter) append() bool {
 	return f.isAppend
 }
 
-func (f *plainHeaderFormatter) format(_ types.HeaderMap, _ types.RequestInfo) string {
+func (f *plainHeaderFormatter) format(_ context.Context) string {
 	return f.staticValue
 }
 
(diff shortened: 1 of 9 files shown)


MengJiapeng:
以下为自动回复    您好，您的邮件我已收到，我会尽快给您回复！

pkg/router/headerformatter.go
```diff
@@ -47,6 +66,73 @@ func (f *plainHeaderFormatter) append() bool {
...
+	} else if strings.HasPrefix(v.variableName, types.VarPrefixRespHeader) {
+		prefix = types.VarPrefixRespHeader
+		isPrefix = true
+	}
+	if isPrefix {
+		headerName := strings.TrimPrefix(v.variableName, prefix)
+		ret, _ := headers.Get(headerName)
+		return ret
+	}
+	switch v.variableName {
```
taoyuanyuan, line 95: we should call `variable.GetString(ctx, variableName)`, to support any variable.

pkg/router/types.go
```diff
@@ -50,7 +50,7 @@ var (
 )
 
 type headerFormatter interface {
-	format(requestInfo api.RequestInfo) string
+	format(headers types.HeaderMap, requestInfo api.RequestInfo) string
```
taoyuanyuan, line 53: I think `headers` and `requestInfo` are not needed,  just `ctx` to support variable..

pkg/router/headerformatter.go
```diff
@@ -47,6 +66,73 @@ func (f *plainHeaderFormatter) append() bool {
...
+	)
+	if strings.HasPrefix(v.variableName, types.VarPrefixReqHeader) {
+		prefix = types.VarPrefixReqHeader
+		isPrefix = true
+	} else if strings.HasPrefix(v.variableName, types.VarPrefixRespHeader) {
+		prefix = types.VarPrefixRespHeader
+		isPrefix = true
+	}
+	if isPrefix {
+		headerName := strings.TrimPrefix(v.variableName, prefix)
```
doujiang24, line 91: maybe we should move these lines to `getHeaderFormatter`? for better performance.

pkg/router/headerformatter.go
MengJiapeng: got it! maybe I should understand how variable works first.

pkg/router/types.go
MengJiapeng: It seems that we didn't set the `requestInfo` into the `ctx`. And we set the request headers and response headers with different context key, but in the `format` method, we can't determine which header we are handing now,  so I think the parameters are necessary here.

pkg/router/headerformatter.go
taoyuanyuan: The variables are global in MOSN, we can get them in different modules of a request lifecycle.

pkg/router/headerformatter.go
MengJiapeng: yeah, I will add a new headerformatter to deal with this situation.

pkg/router/types.go
MengJiapeng: It seems that with `variable` api, we can get any value with `ctx`, I will try later!

pkg/router/headerformatter.go
```diff
@@ -18,17 +18,36 @@
...
 
 	"mosn.io/mosn/pkg/log"
-	"mosn.io/mosn/pkg/types"
 )
 
+var variableValueRegexp = regexp.MustCompile(`^%.+%$`)
+
 func getHeaderFormatter(value string, append bool) headerFormatter {
-	// TODO: variable headers would be support very soon
+	if variableValueRegexp.MatchString(value) {
```
taoyuanyuan, line 32: maybe we don't need introduce regex, and it's easy to judge that first and last string are `%`.

MengJiapeng: Benchmark shows that `strings.HasPrefix` and `strings.HasSuffix` are faster than regex......I will modify it later.
Another question is that if we got a value which not begins and ends with `%`, but contains '%', should we ignore it and return nil or treat it as a plain value?

taoyuanyuan: I think we should treat it as a plain value, beacause we do not prohibit `%` as a part of header

pkg/router/headerformatter.go
```diff
@@ -18,19 +18,21 @@
...
 )
 
 func getHeaderFormatter(value string, append bool) headerFormatter {
-	// TODO: variable headers would be support very soon
-	if strings.Index(value, "%") != -1 {
-		if log.DefaultLogger.GetLogLevel() >= log.WARN {
-			log.DefaultLogger.Warnf("variable headers not support yet, skip, value: %s", value)
+	if len(value) > 2 && strings.HasPrefix(value, "%") && strings.HasSuffix(value, "%") {
+		variableName := strings.Trim(value, "%")
+		// todo cache the variable so we don't need to find it in format method
```
taoyuanyuan, line 29: delete expired  todo comment.

MengJiapeng: todo comment here is not expired. A better implementation can be that we save the `variable.Check` returned `variable` into the variableHeaderFormatter, so we can just get the value in `format` method. But currently, the `variable` api dosen't provide a proper method to get value from variable, so we use `variable.GetString` method which find value from all registered variables , which can be optimized.

taoyuanyuan: oh～ Good idea！

taoyuanyuan:
thanks for your contribution～

taoyuanyuan (approved):

doujiang24 (approved):

PR #1958, opened 2022-02-11
refactor: sds secret manager support a certificate to create different tls config

nejisama:
#1938


pkg/mtls/tls_context_manager.go
```diff
@@ -48,7 +48,7 @@ func NewTLSServerContextManager(cfg *v2.Listener) (types.TLSContextManager, erro
 	}
 	for _, c := range cfg.FilterChains {
 		for _, tlsCfg := range c.TLSContexts {
-			provider, err := NewProvider(&tlsCfg)
+			provider, err := NewProvider(cfg.Name, &tlsCfg)
```
wangfakang, line 51: If the cluster name is the same as the listener name, will this cause tls configuration override issues?

image

nejisama: yes, same index will override. if some one call NewProvider directly, maybe casue override too.
maybe we can add a prefix for tls context manager

wangfakang (approved):
LGTM

doujiang24 (approved):

PR #1922, opened 2022-01-10
feature: introduce a new XProtocol interface GracefulShutdown for graceful stop existing connections.

doujiang24:
related PR:
https://github.com/mosn/api/pull/44
https://github.com/mosn/pkg/pull/59

PR #1967, opened 2022-02-25
optimize: close the connection when there are no active requests after the connpool has been shutdown.

doujiang24:
we implemented the OnGoAway method and call it in pool shutdown.

we can reuse it for graceful shutdown client connection in https://github.com/mosn/mosn/pull/1922


pkg/stream/xprotocol/connpool_multiplex.go
```diff
@@ -166,6 +170,12 @@ func (p *poolMultiplex) NewStream(ctx context.Context, receiver types.StreamRece
 	}
 
 	activeClient := client.(*activeClientMultiplex)
+
+	if atomic.CompareAndSwapUint32(&activeClient.state, GoAway, Connecting) {
+		p.init(activeClient, subProtocol, int(clientIdx))
```
nejisama, line 175: 这里为什么是去重连呢

Issue #1969, opened 2022-03-02
Add weigth != 0 Check for Weighted Round Robin LoadBalancer


antJack:
When adding item to edfSchduler, we do not check weight != 0, which may causes undefined behavior for MOSN
![image](https://user-images.githubusercontent.com/52443884/156288563-2ced1ff5-dd0b-4d88-823f-17d83f6bbb09.png)

This issue proposes the adding of zero check for MOSN LB, not only for WRR

ref:
![image](https://user-images.githubusercontent.com/52443884/156288688-e666654e-f25e-4f1e-9fd5-2685711ffc12.png)

cc. @alpha-baby @jizhuozhi

alpha-baby:
i have another question:

![image](https://user-images.githubusercontent.com/47237632/156298403-58ea706b-213d-4ad4-9973-f0c5123aea19.png)

When `weightFunc` return 0, do we need to remove the iterm from the edf heap?

jizhuozhi:
Which condition would we get 0 weight, and if there is only one item, what will be happened if we remove it?

jizhuozhi:
I think it's necessary to keep the size of items consistent @alpha-baby @antJack

antJack:
what if rounding any weight<=0 to weight=1?

envoy's load_balancing_weight is uint32 and min value is 1.

ref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#config-endpoint-v3-lbendpoint

jizhuozhi:
Maybe it's not a bug but a feature~~.
We could make it clear in the documentation that we don't accept negative numbers (or negative numbers have the highest priority), and 0 is a fallback, 0 items will not be used until all non-zero items are invalid

antJack:
1. `weight=0` will yield `1.0/weight = +Inf` which causes undefined behavior for further code execution. And that's what we should prevent.
2. as for the meaning of `weight<=0`, still cannot imagine in which scenario that we should allow `weight<=0`. Currently I suggest keeping it simple, or there's any scenario?

antJack:
MOSN has host weight const： in pkg/configmanager/parser.go
![image](https://user-images.githubusercontent.com/52443884/156302928-66dfcc3e-d38e-4152-b8b6-197b9c5a3542.png)

jizhuozhi:
So there is only one problem: what should we do when weightFunc return 0 or less than 0

I think this value can be fixed using a consistent contract

[...]

PR #1970, opened 2022-03-02
fix: undefined behavior when weight out of range

jizhuozhi:

#1969

### DIFF
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -18,10 +18,9 @@
 package cluster
 
 import (
+	"mosn.io/mosn/pkg/config/v2"
 	"sync"
 	"time"
-
-	"mosn.io/mosn/pkg/configmanager"
 )
 
 type edfSchduler struct {
@@ -82,11 +81,11 @@ func (edf *edfSchduler) NextAndPush(weightFunc func(item WeightItem) float64) in
 }
 
 func edfFixedWeight(weight float64) float64 {
-	if weight <= float64(configmanager.MinHostWeight) {
-		return float64(configmanager.MinHostWeight)
+	if weight <= float64(v2.MinHostWeight) {
+		return float64(v2.MinHostWeight)
 	}
-	if weight >= float64(configmanager.MaxHostWeight) {
-		return float64(configmanager.MaxHostWeight)
+	if weight >= float64(v2.MaxHostWeight) {
+		return float64(v2.MaxHostWeight)
 	}
 	return weight
 }
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -18,6 +18,7 @@
 package cluster
 
 import (
+	"mosn.io/mosn/pkg/configmanager"
 	"sync"
 	"time"
 )
@@ -52,7 +53,7 @@ func (edf *edfSchduler) Add(item WeightItem, weight float64) {
 	defer edf.lock.Unlock()
 	entry := edfEntry{
 		deadline:   edf.currentTime + 1.0/weight,
-		weight:     weight,
+		weight:     edfFixedWeight(weight),
 		item:       item,
 		queuedTime: time.Now(),
 	}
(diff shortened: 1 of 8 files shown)


pkg/upstream/cluster/edf.go
```diff
@@ -18,6 +18,7 @@
 package cluster
 
 import (
+	"mosn.io/mosn/pkg/configmanager"
```
antJack, line 21: maybe we can move these const to v2 package?

jizhuozhi: I can't make up my mind about this

antJack: just move these to v2/upstream.go
image

alpha-baby: @jizhuozhi you should fallow MOSN import code style

example:

```go
import (
        // build-in package
	"sync"
        // empty line
        // other package
	envoy_api_v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
)
```

pkg/upstream/cluster/edf_test.go
```diff
@@ -18,6 +18,8 @@
 package cluster
 
 import (
+	"math"
+	"mosn.io/mosn/pkg/configmanager"
```
alpha-baby, line 22: should:

```go
import (
	"math"
	"strconv"
	"testing"

	"github.com/stretchr/testify/assert"
	"mosn.io/mosn/pkg/configmanager"
	"mosn.io/mosn/pkg/types"
)
```

pkg/upstream/cluster/edf.go
alpha-baby: > move these to pkg/config/v2, since they are common const value.
> 
> ```go
> const (
> 	MinHostWeight               = uint32(1)
> 	MaxHostWeight               = uint32(128)
> 	DefaultMaxRequestPerConn    = uint32(1024)
> 	DefaultConnBufferLimitBytes = uint32(16 * 1024)
> )
> ```

i agree with you

doujiang24 (approved):

pkg/upstream/cluster/edf.go
antJack: Do we have another PR now?

jizhuozhi: No, this configuration is used in more than one place, but if I submit a revised PR now, it will conflict with this PR

antJack: There're two ways to solve this problem:
1. do all thing in this pr, it's ok, I guest that the amount of affected code won't be too much.
2. a new pr just to move those const define, and rebase the current pr to that new pr, seems little subtle.

pkg/upstream/cluster/edf.go
```diff
@@ -18,6 +18,7 @@
 package cluster
 
 import (
+	"mosn.io/mosn/pkg/config/v2"
```
antJack, line 21: goimports file, should be:
```go
import (
    "sync"
    "time"

    "mosn.io/mosn/pkg/config/v2"
)
```

pkg/upstream/cluster/edf.go
```diff
@@ -52,7 +53,7 @@ func (edf *edfSchduler) Add(item WeightItem, weight float64) {
 	defer edf.lock.Unlock()
 	entry := edfEntry{
 		deadline:   edf.currentTime + 1.0/weight,
-		weight:     weight,
+		weight:     edfFixedWeight(weight),
```
antJack, line 56: should do fix weight before line 55

jizhuozhi:
@alpha-baby `edfFixedWeight` has no race condition, I don't think it should be in critical section.
.

doujiang24:
@jizhuozhi please check the failed test case, thanks.
https://github.com/mosn/mosn/runs/5443770774?check_suite_focus=true#step:4:4799

jizhuozhi:
> @jizhuozhi please check the failed test case, thanks. https://github.com/mosn/mosn/runs/5443770774?check_suite_focus=true#step:4:4799

@doujiang24 Of course, but seems an old test cases covering this bug. I am thinking should I remove it? @antJack 

https://github.com/mosn/mosn/blob/326c1537c1da0acffa3ab10312e408c8ae8243a5/pkg/upstream/cluster/loadbalancer_test.go#L170-L172

https://github.com/mosn/mosn/blob/326c1537c1da0acffa3ab10312e408c8ae8243a5/pkg/upstream/cluster/loadbalancer_test.go#L100-L107

jizhuozhi:
@doujiang24 @antJack @alpha-baby The failed test case was fixed

antJack (approved):
LGTM

doujiang24 (approved):

Issue #1943, opened 2022-01-19
Support slow start mode for WRR.


wangfakang:
### The New Feature

slow start mode is a config of cluster for setting in MOSN to progressively increase amount of traffic for newly added upstream endpoints.

### Your scenes

Describe your use scenes (why need this feature).

### Your advice

Describe the advice or solution of this new feature.

### Environment

- MOSN Version

jizhuozhi:
Is it because some resources need to be warmed up? Such as Java's JIT

jizhuozhi:
If we want to support this function, the more general should be to support the dynamic weight function, for example, the load balancer supports automatic weight reduction and automatic recovery after failure

wangfakang:
> Is it because some resources need to be warmed up? Such as Java's JIT

yes.

wangfakang:
> If we want to support this function, the more general should be to support the dynamic weight function, for example, the load balancer supports automatic weight reduction and automatic recovery after failure

You're right. If you're interested it, welcome to contribute PR.

jizhuozhi:
> > If we want to support this function, the more general should be to support the dynamic weight function, for example, the load balancer supports automatic weight reduction and automatic recovery after failure
> 
> You're right. If you're interested it, welcome to contribute PR.

I'll try to understand how MOSN works and evaluate the cost, if possible, I'll implement this feature after Chinese New Year

jizhuozhi:
Do we need to slow start according to time? @wangfakang

wangfakang:
> Do we need to slow start according to time? @wangfakang For example, we have the slow_start_window as 10s, and the full weight is 100, then each 1s we warmup 10 weight

enh, using only the time dimension is similar to Envoy‘s [slow-start](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start). It may cause the slow-start does not work properly in individual scenarios. For example, when the request arrives, the slow start window has just been restored, as follows:

![image](https://user-images.githubusercontent.com/8215757/150940737-6b9e16ae-f1ee-4014-b7ef-2ac858f7b98d.png)

So, we should add another request dimension, and then control the weight recovery together with the time window. as follows:
![image](https://user-images.githubusercontent.com/8215757/150944610-e6df824d-8279-4291-8867-0ea5a311a686.png)

jizhuozhi:
@wangfakang I don't understand what you mean in graphs.

[...]

PR #2028, opened 2022-05-04
feat: support slow start mode in loadbalancers

jizhuozhi:

#1943 

### Solutions
Use the slow start-related parameters when calculating the weights.
Could see envoy https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start.

**This Pull Request is still in Draft status, some details need to be discussed**

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


jizhuozhi:
@wangfakang Hello, PLAT

PR #2053, opened 2022-06-11
optimize certificate matching logic

MengJiapeng:
#2052 

### Solutions
use first ALPN matched provider only if none of the providers can match the servername

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


taoyuanyuan (approved):
LGTM

nejisama (approved):

PR #2077, opened 2022-07-25
maglev loadbalancer support skipping last chosen host when retrying 

dengqian:

If host is in healthy status, but actually not working ( not deactivated yet), the same host chose when retrying


doujiang24 (dismissed):
LGTM, but I'm not sure if it will break backward compatibility.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -429,10 +429,27 @@ func (lb *maglevLoadBalancer) ChooseHost(ctx types.LoadBalancerContext) types.Ho
...
 	chosen := lb.hosts.Get(index)
+	total := lb.hosts.Size()
+
+	// if retry, means request to last chose host failed, do not use it again
+	context := ctx.DownstreamContext()
+	if ind, err := variable.GetString(context, VarProxyUpstreamIndex); err == nil {
+		if i, err := strconv.Atoi(ind); err == nil {
+			index = i + 1
+		}
+		for i := 0; i < total; i++ {
```
taoyuanyuan, line 440: maybe we could call `chooseHostFromHostList(i)`.

taoyuanyuan: we can directly go to fallback if retry.

dengqian: > we can directly go to fallback if retry.

if we use chooseHostFromHostList, consider about this situation:
```hosts{0:unheathy, 1:healthy but actually unhealthy, 2:healthy but actually unhealthy, ... }```, first time hosts[1] is chosen and then hosts[2], and then hosts[1] again, other hosts will never be chosen when retrying

File: pkg/upstream/cluster/loadbalancer.go [hunk 2]
```diff
@@ -430,9 +430,19 @@ func (lb *maglevLoadBalancer) ChooseHost(ctx types.LoadBalancerContext) types.Ho
...
+			index = i + 1
+		}
+		chosen, index = lb.chooseHostFromHostList(index)
+		variable.SetString(context, VarProxyUpstreamIndex, strconv.Itoa(index))
+	}
+
 	// fallback
 	if !chosen.Health() {
-		chosen = lb.chooseHostFromHostList(index)
+		chosen, index = lb.chooseHostFromHostList(index)
```
taoyuanyuan, line 445: maybe panic if chose is nil.

taoyuanyuan: index+1

taoyuanyuan: We can optimize the logic by calling  `chooseHostFromHostList` only once.

taoyuanyuan (approved):
LGTM

doujiang24 (approved):

PR #2080, opened 2022-07-26
bugfix: tcp proxy handle write timeout or write close event

dengqian:
In some case, upstream connection write failed and then upstream connection is closed, but streamproxy didn't handle these write events, but downstream connection is alive. So data is only read one hand, not proxying to upstream

### DIFF
--- a/pkg/filter/network/streamproxy/streamproxy.go
+++ b/pkg/filter/network/streamproxy/streamproxy.go
@@ -81,10 +81,7 @@ func (p *proxy) OnData(buffer buffer.IoBuffer) api.FilterStatus {
 	bytesRecved := p.requestInfo.BytesReceived() + uint64(buffer.Len())
 	p.requestInfo.SetBytesReceived(bytesRecved)
 
-	err := p.upstreamConnection.Write(buffer.Clone())
-	if err == types.ErrConnectionHasClosed {
-		p.onUpstreamEvent(api.LocalClose)
-	}
+	p.upstreamConnection.Write(buffer.Clone())
 	buffer.Drain(buffer.Len())
 	return api.Stop
 }
@@ -117,6 +114,8 @@ func (p *proxy) getUpstreamConnection(ctx types.LoadBalancerContext, snapshot ty
 	}
 }
 
+const defaultConnectRetryTimes = 3
+
 func (p *proxy) initializeUpstreamConnection() api.FilterStatus {
 	clusterName := p.getUpstreamCluster()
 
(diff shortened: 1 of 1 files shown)


pkg/filter/network/streamproxy/streamproxy.go
```diff
@@ -139,31 +145,46 @@ func (p *proxy) initializeUpstreamConnection() api.FilterStatus {
...
 		cluster: clusterInfo,
 	}
-	connectionData := p.getUpstreamConnection(ctx, clusterSnapshot)
-	if connectionData.Connection == nil {
-		p.requestInfo.SetResponseFlag(api.NoHealthyUpstream)
-		p.onInitFailure(NoHealthyUpstream)
 
-		return api.Stop
+	retryTime := clusterSnapshot.HostSet().Size()
+	if retryTime > 3 {
```
doujiang24, line 150: better use a const variable instead of hardcode number 3? make the number readable.

doujiang24 (approved):

pkg/filter/network/streamproxy/streamproxy.go
```diff
@@ -74,12 +75,16 @@ func NewProxy(ctx context.Context, config *v2.StreamProxy, net string) Proxy {
...
-		log.DefaultLogger.Debugf("[%s proxy] [ondata] read data , len = %v", p.network, buffer.Len())
+		log.DefaultLogger.Debugf("[%s proxy] [ondata] read data , len = %v, local addr:%s, upstream addr:%s",
+			p.network, buffer.Len(), p.readCallbacks.Connection().LocalAddr().String(), p.upstreamConnection.RemoteAddr().String())
 	}
 	bytesRecved := p.requestInfo.BytesReceived() + uint64(buffer.Len())
 	p.requestInfo.SetBytesReceived(bytesRecved)
 
-	p.upstreamConnection.Write(buffer.Clone())
+	err := p.upstreamConnection.Write(buffer.Clone())
+	if err == types.ErrConnectionHasClosed {
```
taoyuanyuan, line 85: no need to call onUpstreamEvent here, because Write error would call onUpstreamEvent directly.

pkg/filter/network/streamproxy/streamproxy.go
```diff
@@ -193,7 +214,7 @@ func (p *proxy) onUpstreamData(buffer types.IoBuffer) {
 
 func (p *proxy) onUpstreamEvent(event api.ConnectionEvent) {
 	switch event {
-	case api.RemoteClose:
+	case api.RemoteClose, api.OnWriteTimeout, api.OnWriteErrClose:
```
taoyuanyuan, line 217: maybe we should do it in `onDownstreamEvent` also.

taoyuanyuan (approved):
LGTM

PR #2082, opened 2022-07-26
[bugfix] add connection event listener after ac.element set to avoid panic

dengqian:
In some case, upstream connection can be closed very quickly by another go routine, so ```AddConnectionEventListener``` should be called after ```ac.element = e```,  in case close event come before ac.element  is set.

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -534,9 +534,20 @@ func (al *activeListener) OnNewConnection(ctx context.Context, conn api.Connecti
 		return
 	}
 	ac := newActiveConnection(al, conn)
+	// connection may be closed by upstream before added to conn list
+	if conn.State() != api.ConnClosed {
+		return
+	}
+
+	if conn.LocalAddr().Network() == "udp" {
+		network.SetUDPProxyMap(network.GetProxyMapKey(conn.LocalAddr().String(), conn.RemoteAddr().String()), conn)
+	}
 
 	e := al.conns.PushBack(ac)
 	ac.element = e
+
+	// add event listener after setting ac.element to avoid delete empty element
+	ac.conn.AddConnectionEventListener(ac)
 
 	atomic.AddInt64(&al.handler.numConnections, 1)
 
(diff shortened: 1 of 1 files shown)


doujiang24:
Nice catch. But I still have two questions:
1. In which scenario, `closed very quickly by another goroutine` will happen?
2. please help to paste a piece of the panic log, I haven't figured out the panic backtrace.

pkg/server/handler.go
```diff
@@ -534,19 +534,27 @@ func (al *activeListener) OnNewConnection(ctx context.Context, conn api.Connecti
...
+	}
+
+	if conn.LocalAddr().Network() == "udp" {
+		network.SetUDPProxyMap(network.GetProxyMapKey(conn.LocalAddr().String(), conn.RemoteAddr().String()), conn)
+	}
 
 	e := al.conns.PushBack(ac)
 	ac.element = e
 
+	// add event listener after setting ac.element to avoid delete empty element
```
doujiang24, line 549: Do you mean panic here?
https://github.com/mosn/mosn/blob/aab8c094446abf98652b6b68a798ba553dff0dad/pkg/server/handler.go#L640

dengqian: Yes, that's what I mean. If event listener added before ```ac.element``` set, ``` al.conns.Remove(ac.element) ``` can be called ealier

doujiang24: got it, thanks!

doujiang24 (approved):

nejisama (approved):

pkg/server/handler.go
```diff
@@ -534,19 +534,27 @@ func (al *activeListener) OnNewConnection(ctx context.Context, conn api.Connecti
...
+	ac.conn.AddConnectionEventListener(ac)
```
taoyuanyuan, line 550: There is another issue that `ac` would not be removed if conn has closed at this time.

taoyuanyuan:
maybe we should call `newActiveConnection` before `InitializeReadFilters`.

PR #2085, opened 2022-07-28
optimize: reduce memory usage during http post requests

taoyuanyuan:

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/stream/http/stream.go
```diff
@@ -197,10 +197,7 @@ func (conn *streamConnection) Read(p []byte) (n int, err error) {
 func (conn *streamConnection) Write(p []byte) (n int, err error) {
 	n = len(p)
 
-	// TODO avoid copy
-	buf := buffer.GetIoBuffer(n)
-	buf.Write(p)
-
+	buf := buffer.NewIoBufferBytes(p)
```
wangfakang, line 200: could remove the comment in the `note' section.
https://github.com/mosn/mosn/blob/2a191b6387bc81f7928237ccce3a552cfa586e5e/pkg/stream/http/stream.go#L825-L827

nejisama (approved):

doujiang24 (approved):

PR #2092, opened 2022-08-02
optimize host update logic

dengqian:
when new hosts is subset of old hosts and all new hosts unhealthy, skip updating

### DIFF
--- a/pkg/upstream/cluster/cluster_manager.go
+++ b/pkg/upstream/cluster/cluster_manager.go
@@ -172,7 +172,7 @@ func (cm *clusterManager) AddOrUpdateClusterAndHost(cluster v2.Cluster, hostConf
 	return cm.UpdateCluster(cluster, func(oc, nc types.Cluster) {
 		UpdateClusterResourceManagerHandler(oc, nc)
 		CleanOldClusterHandler(oc, nc)
-		NewSimpleHostHandler(nc, hostConfigs)
+		NewHealthySimpleHostHandler(nc, hostConfigs)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


pkg/upstream/cluster/cluster_manager.go
```diff
@@ -250,15 +250,31 @@ func (cm *clusterManager) RemovePrimaryCluster(clusterNames ...string) error {
 	return nil
 }
 
+
 // EDS Handler
-func NewSimpleHostHandler(c types.Cluster, hostConfigs []v2.Host) {
+func NewHealthySimpleHostHandler(c types.Cluster, hostConfigs []v2.Host) {
 	snap := c.Snapshot()
 	hosts := make([]types.Host, 0, len(hostConfigs))
+	allHostDown := true
```
doujiang24, line 258: maybe `needUpdate` could be better? since it is not only for healthy.

pkg/upstream/cluster/cluster_manager.go
```diff
@@ -250,15 +250,31 @@ func (cm *clusterManager) RemovePrimaryCluster(clusterNames ...string) error {
...
+	snap.HostSet().Range(func(host types.Host) bool {
+		oldHosts[host.AddressString()] = true
+		return true
+	})
+
 	for _, hc := range hostConfigs {
-		hosts = append(hosts, NewSimpleHost(hc, snap.ClusterInfo()))
+		host := NewSimpleHost(hc, snap.ClusterInfo())
+		hosts = append(hosts, host)
+		if host.Health() || oldHosts[host.AddressString()] != true {
```
doujiang24, line 269: I'm not sure this is the right way, two questions:
1. skip updating when it is subset of the old cluster, then, how to remove a host from a old cluster?
2. host also contains some other information, eg. metadata, but we just checking the address.

nejisama: > I'm not sure this is the right way, two questions:
> 
> 1. skip updating when it is subset of the old cluster, then, how to remove a host from a old cluster?
> 2. host also contains some other information, eg. metadata, but we just checking the address.


i think we should not skip. 
Although the host to be updated is unhealthy, the host to be deleted may no longer belong to the cluster

dengqian: > > I'm not sure this is the right way, two questions:
> > 
> > 1. skip updating when it is subset of the old cluster, then, how to remove a host from a old cluster?
> > 2. host also contains some other information, eg. metadata, but we just checking the address.
> 
> i think we should not skip. Although the host to be updated is unhealthy, the host to be deleted may no longer belong to the cluster
I agree with you. The problem I have encountered is that when the number of pods is relatively small, some pod status is not updated because the associated node is down , and the actually available pods status may jitter(network instability in edge-scenario between pods and api-server).

dengqian: > > I'm not sure this is the right way, two questions:
> > 
> > 1. skip updating when it is subset of the old cluster, then, how to remove a host from a old cluster?
> > 2. host also contains some other information, eg. metadata, but we just checking the address.
> 
> i think we should not skip. Although the host to be updated is unhealthy, the host to be deleted may no longer belong to the cluster

I agree with you. The problem I have encountered is that when the number of pods is relatively small, some pod status is not updated because the associated node is down , and the actually available pods status may jitter(network instability in edge-scenario between pods and api-server).

PR #2136, opened 2022-09-29
bugfix: return error when get cookie value of not-exist-key

greedying:

no issues associated

### Solutions

when get a cookie value of a key, if the key not exists, return `variable.ValueNotFound` and an error 

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/stream/http/var.go
+++ b/pkg/stream/http/var.go
@@ -19,6 +19,7 @@ package http
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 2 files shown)


taoyuanyuan:
thanks for your contribution！

taoyuanyuan (approved):

antJack (approved):
LGTM

PR #2160, opened 2022-10-31
Builds with Go 1.18, but tests with 1.17, 1.18 and 1.19

### LINKED ISSUE #2153: Golang update to 1.18
### The New Feature

Describe the new feature you want to support clearly.

### Your scenes

Describe your use scenes (why need this feature).

### Your advice

Describe the advice or solution of this new feature.

### Environment

- MOSN Version

codefromthecrypt:
This uses a test matrix to ensure the floor version of Mosn is two
behind latest Go, even if one behind latest (currently 1.18) is used to
build binaries.

Fixes #2153


pkg/admin/debug/debug_api.go
```diff
@@ -1,4 +1,4 @@
-// +build mosn_debug
+//go:build mosn_debug
```
antJack, line 1: keep both build constraints, otherwise go<1.17 may fails.
ref: 
- https://github.com/golang/sys/blob/master/cpu/cpu_gccgo_x86.go#L5
- https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives

pkg/module/http2/transport.go
```diff
@@ -2604,3 +2604,20 @@ func traceFirstResponseByte(trace *httptrace.ClientTrace) {
 		trace.GotFirstResponseByte()
 	}
 }
+
+func traceHasWroteHeaderField(trace *httptrace.ClientTrace) bool {
```
antJack, line 2608: why delete go111.go and not_go111.go? seems that we should still keep these files? 
ref: https://cs.opensource.google/go/x/net/+/master:http2/go111.go

pkg/types/config_test.go
```diff
@@ -1,4 +1,4 @@
-// +build linux darwin
+//go:build linux
```
antJack, line 1: should keep darwin?

pkg/admin/debug/debug_api.go
codefromthecrypt: gotcha I didn't realize go 1.18 also meant keep go 1.17 (which is no longer supported)

I will revert the changes. Please note that code using 1.18 `any` will also fail on 1.17

pkg/module/http2/transport.go
codefromthecrypt: do you want to compile in a version less than go 1.11? I deleted <1.11 support which those files were required for...

pkg/types/config_test.go
codefromthecrypt: accident

codefromthecrypt:
@antJack can you clarify what minimum go version you would like?

Right now, Go supports 1.18 and 1.19 and there is some impact.

For example, we have code that tests to see if it is at least go 1.11. That seems surely too old version to worry about.

Another place we test about supported TLS versions that 1.0 and 1.1 rejects, but it rejects earlier with Go 1.18 ex.
```
=== RUN   TestServer_Rejects_TLS10
    server_test.go:152: tls: no supported versions satisfy MinVersion and MaxVersion
```

Finally, //go:build directive was added in go 1.17, not 1.18


So, my thinking when @taoyuanyuan raised #2153 was that the intent was to move to the floor version supported by go. Not, go 1.11 or 1.16, but 1.18, and not make special cases to make 1.16 work (ex avoid `//go:build`)

With some clarity I can proceed on some changes here, just let me know!

codefromthecrypt:
I made minimal changes except 

```
=== RUN   TestServer_Rejects_TLS10
    server_test.go:152: tls: no supported versions satisfy MinVersion and MaxVersion
```

maybe I can do some surgery if we expect to want to have to test this before go 1.18 still. It would be nicer if we don't and then we can just remove the tests. lemme know.

antJack:
go1.18 is our suggested version, however we should also provide backward compatibility as much as possible, since we cannot force everyone to use 1.18, especially for some local builds.
another simple reason for my comments is that the official go also does these backward compatibility as:
https://github.com/golang/sys/blob/master/cpu/cpu_gccgo_x86.go#L5
https://cs.opensource.google/go/x/net/+/master:http2/go111.go

codefromthecrypt:
@antJack if any library uses `any` instead of `interface{}` no one below 1.18 will be able to compile. How do you suggest implementing a "not break" policy without testing in a matrix? and if we don't can we at least suggest what the floor version is? I think we'd agree 1.11 is too low, so what isn't too low?

codefromthecrypt:
fwiw the main reason I did all the work in mosn recently was to help set the floor version to 1.18. If we can't even do that, maybe I need to close the wazero related PR because surely by next year wazero will not support old versions of go, and I really doubt many will. I think it is critical to think through this and try to balance concern for the past with the present and maybe future.

pkg/module/http2/transport.go
codefromthecrypt: I left removing this for another person. It isn't critical to do in this PR

codefromthecrypt:
ok the PR is synced with latest policy. hopefully tests pass but if don't I'll look at the failure.

codefromthecrypt:
ok this is using current version policy as well updated pkg and api. please merge @taoyuanyuan as I have run out of time a while ago on remedial maintenance of mosn

antJack (approved):
LGTM

taoyuanyuan (approved):
LGTM

PR #2164, opened 2022-11-03
Adds integrated benchmarks for proxywasm

codefromthecrypt:
This adds benchmarks for the only proxy-wasm compatible ABI available in mosn: v1. This is mainly an effort to help folks recognize the performance overhead and also how critical being incompatible is for the ecosystem.

For example, the wasm in the below test only adds one header and nothing else. To do that takes over 60us. In comparison, other wasm handlers can do the same in single digit us even 2 or 3.
```
Benchmark_BaseCase-16       	    8696	    133748 ns/op
Benchmark_ProxyWasmV1-16    	    5760	    195353 ns/op
```

cc @Taction @taoyuanyuan @antJack


antJack (approved):
LGTM

taoyuanyuan (approved):
LGTM

PR #2168, opened 2022-11-04
work-in-progress: Add http-wasm stream filter

codefromthecrypt:
This adds a new httpwasm stream filter based on the [http-wasm][1] Application Binary Interface (ABI), specifically using the [http-handler][2] ABI via [http-wasm-host-go][3]. The net result is a simple and fast way for writing middleware that mosn can load and unload dynamically. Currently, the only guest language is [TinyGo][4], which the examples are written in, but others can be requested [here][1].

Here's an example custom handler written in TinyGo which acts as a simple router.
```go
// handleRequest implements a simple HTTP router.
func handleRequest(req api.Request, resp api.Response) (next bool, reqCtx uint32) {
	// If the URI starts with /host, trim it and dispatch to the next handler.
	if uri := req.GetURI(); strings.HasPrefix(uri, "/host") {
		req.SetURI(uri[5:])
		next = true // proceed to the next handler on the host.
		return
	}

	// Serve a static response
	resp.Headers().Set("Content-Type", "text/plain")
	resp.Body().WriteString("hello")
	return // skip the next handler, as we wrote a response.
}
```

[1]: https://http-wasm.io
[2]: https://http-wasm.io/http-handler-abi/
[3]: https://github.com/http-wasm/http-wasm-host-go
[4]: https://github.com/http-wasm/http-wasm-guest-tinygo


codefromthecrypt:
cc @antJack @taoyuanyuan @jcchavezs @anuraaga I setup a shared branch but I suspect the main thing to do here is get the configuration and naming etc into a shape that makes sense for mosn. I'm not sure what that is. I'll be away on travel a while, just raised this to get things off my computer and brain in case others can move things forward.

pkg/filter/stream/httpwasm/filter_test.go
```diff
@@ -0,0 +1,353 @@
...
+func TestURI(t *testing.T) {
+	var backend *httptest.Server
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if want, have := "/v1.0/hello?name=teddy", r.URL.RequestURI(); want != have {
+			t.Fatalf("unexpected request URI, want: %q, have: %q", want, have)
+		}
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			t.Fatal(err)
+		}
```
jcchavezs, line 66:
```suggestion
		}
		defer r.Body.Close()
```

pkg/filter/stream/httpwasm/host.go
```diff
@@ -0,0 +1,315 @@
...
+
+func (host) GetURI(ctx context.Context) string {
+	// TODO(anuraaga): There is also variable.GetProtocolResource(ctx, api.URI), unclear if they must be kept in sync.
+	p := mustGetString(ctx, types.VarPath)
+	q, qErr := variable.GetString(ctx, types.VarQueryString)
+	if qErr != nil {
+		// No query, so an error is returned.
+		return p
+	}
+	return fmt.Sprintf("%s?%s", p, q)
```
jcchavezs, line 149: Should you urlencode `q`?

codefromthecrypt: yeah we should clarify that as raw I think in the abi doc, then update the base test with a value that is different if URL encoded. Wanna try this as a warm up change?

https://github.com/http-wasm/http-wasm/blob/main/content/http-handler-abi.md#set_uri
https://github.com/http-wasm/http-wasm-host-go/blob/main/handler/nethttp/host.go#L57-L76
https://github.com/http-wasm/http-wasm-host-go/blob/main/internal/test/testdata/e2e/uri.wat#L17
https://github.com/http-wasm/http-wasm-host-go/blob/main/handler/nethttp/middleware_test.go#L108-L146

codefromthecrypt:
ps I added this because it is quite difficult to interpret lint right now https://github.com/mosn/mosn/issues/2187

codefromthecrypt:
reminder this is on a shared branch, so anyone in http-wasm org can progress this.

The main thing needed here is to close out the percent thing that was raised out almost 3 weeks ago, and see if there is any drift impact as I know changing type signatures was tossed around. Remember example code here may need to be updated.

Next, the last mile stuff like testing config and adding integration tests such as exist for proxy-wasm. Trailers can be done as a follow-up as we already have a flag to disable them and most users won't process trailers.

Even better would be an actual end user asking for trailers to work in mosn as there's effort required to update everything to latest fasthttp and also [integrate it](https://github.com/mosn/pkg/issues/66). Meanwhile, we should try to get users unblocked by getting this in soon.

codefromthecrypt:
I just updated this to latest wazero, and added an integration test similar to the one for proxy-wasm. There's more work needed, but anyway at least some progress.

codefromthecrypt:
@anuraaga ps if you 👍 this comment, I'll switch your name off the TODOs so they don't become immortal :D

codefromthecrypt:
Added benchmark, showing this has 2.3us overhead to add a header vs proxy-wasm which is 12.4us

```
goos: darwin
goarch: arm64
pkg: mosn.io/mosn/test/integrate/httpwasm
Benchmark_Baseline
Benchmark_Baseline-12    	   19088	     60136 ns/op
Benchmark_HttpWasm
Benchmark_HttpWasm-12    	   19058	     62417 ns/op
PASS
```

codefromthecrypt:
just re-ran the benchmarks with latest wazero, since the overhead is a bit less than prior

codefromthecrypt:
closing this, as I won't personally be completing it, and it is approaching a year old. someone in the http-wasm org can re-open as it is on a shared branch, or start again!

PR #2169, opened 2022-11-06
Fixes proxywasm test which accidentally used v2 imports

codefromthecrypt:
This adds a sanity check to the wasm proxytest which accidentally used the wrong ABI.

cc @taoyuanyuan @antJack


antJack (approved):
LGTM

taoyuanyuan (approved):

PR #2172, opened 2022-11-09
Moves to consolidated proxy-wasm implementation and enables wazero by default

codefromthecrypt:
This migrates to the latest proxy-wasm-go-host which folds in functionality formerly in this repository. This adds wazero by default, which has no CGO dependency and routinely released.

By having a default runtime, more tests can run by default, and mosn proxy-wasm can work on more platforms such as windows.

*Note*: Docker still has wasmer by default like it did before. This means those using docker can choose existing engine of wasmer or wazero. Since this choice is in configuration, those wanting to use wazero need to update their files to do so.

For local development or custom builds, this adds a new `build-local-wasmer` target which lets local users build with wasmer without passing TAGS. For example, I use this to avoid using docker to test things.

See https://github.com/mosn/proxy-wasm-go-host/pull/24


pkg/wasm/runtime/wasmer/wasmer_unsupported.go
```diff
@@ -0,0 +1,21 @@
+//go:build !wasmer
```
codefromthecrypt, line 1: I had to add an empty file here because there are places in mosn which do an underscore import of wasmer. This allows those call sites to stick around even if the build flag isn't there. cc @taoyuanyuan

antJack (dismissed):
LGTM

codefromthecrypt:
we can wait until this and I can do both at the same time, or you can merge this first.. your choice https://github.com/mosn/proxy-wasm-go-host/pull/27

codefromthecrypt:
I will update this since the other repo is now ready

codefromthecrypt:
I found a glitch, integrating this. I have this PR locally complete, but it depends on https://github.com/mosn/proxy-wasm-go-host/pull/28

go.mod
```diff
@@ -131,3 +132,5 @@ require (
 replace github.com/envoyproxy/go-control-plane => github.com/envoyproxy/go-control-plane v0.10.0
 
 replace istio.io/api => istio.io/api v0.0.0-20211103171850-665ed2b92d52
+
+replace mosn.io/proxy-wasm-go-host => ../pp
```
codefromthecrypt, line 136: this is why build is broken for now. I'll migrate it once proxy-wasm-go-host is merged

codefromthecrypt:
@antJack @taoyuanyuan milestone time maybe!

fyi wazero functions used here will be in a tagged release in one week, so I will follow-up when that happens and bump proxy-wasm-go-host then here again. meanwhile we can test using the untagged one.

cmd/mosn/main/mosn.go
```diff
@@ -74,7 +74,7 @@ import (
 	_ "mosn.io/mosn/pkg/upstream/servicediscovery/dubbod"
 	_ "mosn.io/mosn/pkg/wasm/abi/proxywasm010"
 	_ "mosn.io/mosn/pkg/wasm/abi/proxywasm020"
-	_ "mosn.io/mosn/pkg/wasm/runtime/wasmer"
+	_ "mosn.io/mosn/pkg/wasm/runtime/wazero"
```
antJack, line 77: keep both wazero and wasmer, so that users can choose which one to use

test/integrate/proxywasm/proxywasm_wasmer_test.go
```diff
@@ -0,0 +1,17 @@
+//go:build wasmer
+// +build wasmer
+
```
antJack, line 3: add
```
/*
 * Licensed to the Apache Software Foundation (ASF) under one or more
```

test/integrate/proxywasm/proxywasm_wazero_test.go
```diff
@@ -0,0 +1,14 @@
+package wasm_test
```
antJack, line 1: add build tag as same as in `test/integrate/proxywasm/proxywasm_wasmer_test.go`

pkg/wasm/runtime/wazero/wazero.go
```diff
@@ -1,6 +1,3 @@
-//go:build wasmer
-// +build wasmer
-
 /*
```
antJack, line 1: by default, we should let both wazero and wasmer to be built in

codefromthecrypt: wazero doesn't need a build tag because has no deps. but wasmer used to have a build tag because it has system library and portability issues. so you want to have a build tag for wazero knowing this?

cmd/mosn/main/mosn.go
codefromthecrypt: sure but this will keep mosn main portability coupled to what libraries wasmer distributes. can do

antJack: what if we keep import both vms, and add a makefile target like `build-local-wasmer` which build in wamser, while the default one `build-local` just build with wazero?

codefromthecrypt:
@taoyuanyuan do you agree with @antJack that should make wazero disabled by default (by adding a build tag) despite it being compatible with same platforms as normal go? It adds complexity imho because a tree of build-tag dependent sources (notably tests)

codefromthecrypt:
I added wasmer in by default as it was, but I think it is a bad idea to make both wasmer and also wazero build tag conditional. this leaves a lot of tests disabled by default which isn't progress imho.

pkg/wasm/runtime/wazero/wazero.go
codefromthecrypt: having both in main is here now. what you are asking here is to make sure there's no engine available unless a user adds a build tag. I think this is a mistake as it will ensure tests don't run as the fake isn't good enough to test anything. This leaves one question and if you insist I can disable even wazero by default, just be prepared for things to be broke and no one notice.

cmd/mosn/main/mosn.go
codefromthecrypt: so let's clarify, as I think really there are a few things. also keep in mind wasmer 3 just came out in rust and once goes to wasmer-go likely will be incompatible in some ways.

right now build-local adds wazero because it is not tag-specific. Users can already supply TAGS to the makefile to add wasmer, which will result in wazero and wasmer as engine choices. So, basically build-local-wasmer is same as setting TAGS=wasmer

If the desire is to intentionally remove wazero (2MB from the 140+MB binary), then we have to add a build tag to disable wazero. However, doing that forces us to make tests depend only on mock and basically that ensures they won't happen. Also, it provides no out-of-box wasm runtime w/o CGO which I thought was the original intent of the request to add wazero. https://github.com/mosn/proxy-wasm-go-host/issues/15

thoughts?

pkg/wasm/runtime/wazero/wazero.go
antJack: i think there is somewhat misunderstanding, my first thought was to keep both vms built in, not `there's no engine available unless a user adds a build tag`. And we allow to just build in wazero in the default make target `build-local`, but at the same time we should allow user to turn on wasmer by simple way just like adding a new make target `build-local-wasmer`, in other words, we do not wishes that our users are required to do any code modification (include `import`) to turn on wasmer

codefromthecrypt: ok so what's left is user and build instructions.

With current commit the docker build adds by default wasmer, because before it also had the build tag set for it.

for build-local, there's the choice to reduce that to the platform choices of wasmer 1.04, by adding the tag there also.

There's also the choice to show users how to use the existing TAGS variable, and finally to do that for them with a new target which adds that. Which do you prefer?

codefromthecrypt: I guess what I'm saying is.. before build-local didn't build in wasmer as far as I can tell, even if docker did. If you want the change to add wazero to also include build local for wasmer, we can do that. However, before it didn't except when someone set the build tag wasmer via TAGS.

codefromthecrypt: ex in 1.2.0 you can see this, which means wasmer doesn't build unless you added it via TAGS, which docker does and build-local doesn't.

```
pkg/wasm/runtime/wasmer/vm.go
1://go:build wasmer
2:// +build wasmer
```

codefromthecrypt: added it anyway :D

codefromthecrypt:
@antJack PTAL I added the wasmer build-local thing even if it is unrelated to this change. Hope it helps!

For example, you can see it is passing the tags now.
```bash
$ make build-local-wasmer
GO111MODULE=on CGO_ENABLED=1 go build -tags wasmer \
		-ldflags "-B 0x29395398bda831d8e141ccbe5ee66076585cfcc5 -X main.Version=v1.2.0(ad4cb3063) -X mosn.io/mosn/pkg/istio.IstioVersion=1.10.6" \
```

test/integrate/proxywasm/proxywasm_wazero_test.go
codefromthecrypt: I think we agreed to leave this one enabled by default so that tests can run. so dismissing this

codefromthecrypt:
also I fixed the description to reflect new status after this PR

test/integrate/proxywasm/proxywasm_wasmer_test.go
antJack: plz make sure that every files contains those agreements.

codefromthecrypt: found one more.. I think that's it

antJack (approved):
LGTM

taoyuanyuan (approved):
LGTM

Issue #2176, opened 2022-11-11
【BUG】data race in health flag variable


alpha-baby:
### Describe the bug

[A clear and concise description of what the bug is.](https://github.com/mosn/mosn/blob/master/pkg/upstream/cluster/health.go#L46)

### Environment

- MOSN Version

PR #2177, opened 2022-11-11
fix: data race

alpha-baby:

#2176 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/upstream/cluster/health.go
+++ b/pkg/upstream/cluster/health.go
@@ -41,16 +41,24 @@ func SetHealthFlag(p *uint64, flag api.HealthFlag) {
 	if p == nil {
 		return
 	}
-	f := atomic.LoadUint64(p)
-	f |= uint64(flag)
-	atomic.StoreUint64(p, f)
+	for {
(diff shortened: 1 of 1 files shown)


doujiang24:
LGTM. But one question, why there are two health checkers updating the flags concurrency? 
From my common sense, health checkers in concurrency for a single host is not reasonable.

alpha-baby:
> LGTM. But one question, why there are two health checkers updating the flags concurrency? From my common sense, health checkers in concurrency for a single host is not reasonable.

Maybe host object is the same one, but their address is the same one. 

example:

there is two host that address is same. it's in two differet cluster and heathyChcker, so will currently update it's flag.
Of course, the probability of data race is very small.

doujiang24:
> there is two host that address is same. it's in two differet cluster and heathyChcker, so will currently update it's flag. Of course, the probability of data race is very small.

two host objects that have the same address, still are two host objects? there won't be race?
The race could happen, when the same host is shared in two different clusters, right? Could it be possible? @nejisama

PR #2178, opened 2022-11-13
feat: EdfLoadBalancer supports slow start mode

jizhuozhi:

#1943 

### Solutions
Use the slow start-related parameters when calculating the weights.
Could see envoy https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start.

### UT result
```console
2022-11-13 16:37:06,862 [INFO] [network] [ register pool factory] register protocol: mock factory
=== RUN   Test_slowStartHostWeightFunc
--- PASS: Test_slowStartHostWeightFunc (0.00s)
PASS
```

### Benchmark

### Code Style
+ Show `Golint` result


jizhuozhi:
Hello, @alpha-baby and @wangfakang . I'm sorry it's too long to support slow-start, PLAT

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -196,7 +197,11 @@ type WRRLoadBalancer struct {
 
 func newWRRLoadBalancer(info types.ClusterInfo, hosts types.HostSet) types.LoadBalancer {
 	wrrLB := &WRRLoadBalancer{}
-	wrrLB.EdfLoadBalancer = newEdfLoadBalancerLoadBalancer(hosts, wrrLB.unweightChooseHost, wrrLB.hostWeight)
+	if info == nil {
```
alpha-baby, line 200: 这种我感觉不太好，直接把 info 传到 Edf 里面去得了，当然看看其他人是啥想法

pkg/upstream/cluster/cluster_manager.go
```diff
@@ -157,13 +157,33 @@ func InheritClusterHostsHandler(oc, nc types.Cluster) {
 	nc.UpdateHosts(oc.Snapshot().HostSet())
 }
 
+func TransferHostStartTimeHandler(oc, nc types.Cluster) {
+	if oc == nil {
+		return
+	}
+
```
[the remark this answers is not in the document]
jizhuozhi: has any condition that `nc` is `nil`?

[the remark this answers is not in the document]
alpha-baby: 服务发现的时候如果服务有空了， 没有 endpoint 了

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -330,6 +346,29 @@ func newEdfLoadBalancerLoadBalancer(hosts types.HostSet, unWeightChoose func(typ
 	return lb
 }
 
+// slowStartHostWeightFunc progressively increases amount of traffic for newly added upstream hosts
+func slowStartHostWeightFunc(slowStart types.SlowStart, now func() time.Time, hostWeightFunc func(host WeightItem) float64) func(host WeightItem) float64 {
+	return func(host WeightItem) float64 {
```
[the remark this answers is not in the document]
jizhuozhi: see #1943 , it not belongs to this PR.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -330,6 +338,29 @@ func newEdfLoadBalancerLoadBalancer(hosts types.HostSet, unWeightChoose func(typ
...
+		w := hostWeightFunc(host)
+		h, ok := host.(types.Host)
+		if !ok {
+			return w
+		}
+
+		duration := now().Sub(h.StartTime())
+		window := slowStart.SlowStartWindow
+		if duration > window {
```
wangfakang, line 352: performance tips: duration >= window

pkg/upstream/cluster/cluster_manager.go
```diff
@@ -173,6 +202,7 @@ func (cm *clusterManager) AddOrUpdateClusterAndHost(cluster v2.Cluster, hostConf
 		UpdateClusterResourceManagerHandler(oc, nc)
 		CleanOldClusterHandler(oc, nc)
 		NewSimpleHostHandler(nc, hostConfigs)
+		TransferClusterHostStatesHandler(nc, nc)
```
wangfakang, line 205: TransferClusterHostStatesHandler(nc, nc) -> TransferClusterHostStatesHandler(oc, nc)

jizhuozhi:
Hello, @alpha-baby and @wangfakang . All conversations are resolved, PLAT

pkg/upstream/cluster/cluster_manager.go
```diff
@@ -157,13 +157,42 @@ func InheritClusterHostsHandler(oc, nc types.Cluster) {
...
+}
+
 // AddOrUpdatePrimaryCluster will always create a new cluster without the hosts config
 // if the same name cluster is already exists, we will keep the exists hosts.
 func (cm *clusterManager) AddOrUpdatePrimaryCluster(cluster v2.Cluster) error {
 	return cm.UpdateCluster(cluster, func(oc, nc types.Cluster) {
 		UpdateClusterResourceManagerHandler(oc, nc)
 		CleanOldClusterHandler(oc, nc)
 		InheritClusterHostsHandler(oc, nc)
+		TransferClusterHostStatesHandler(oc, nc)
```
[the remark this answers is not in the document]
jizhuozhi: Resolved

pkg/upstream/cluster/cluster.go
```diff
@@ -82,6 +82,32 @@ func NewClusterInfo(clusterConfig v2.Cluster) types.ClusterInfo {
 		info.idleTimeout = clusterConfig.IdleTimeout.Duration
 	}
 
+	// set SlowStart
```
wangfakang, line 85: It only takes effect when the user configures `slowstart`, and shouldn't  enabled by default.

wangfakang: But the value of `info.slowStart.Mode` is never "".
https://github.com/mosn/mosn/blob/ffedd06a6bc259268cfa71ef920a2e2114c38a2f/pkg/upstream/cluster/cluster.go#L85-L91

jizhuozhi: It's my mistake, I will fix it

jizhuozhi: Resolved, PLAT

pkg/upstream/cluster/cluster_manager.go
```diff
@@ -173,6 +201,7 @@ func (cm *clusterManager) AddOrUpdateClusterAndHost(cluster v2.Cluster, hostConf
 		UpdateClusterResourceManagerHandler(oc, nc)
 		CleanOldClusterHandler(oc, nc)
 		NewSimpleHostHandler(nc, hostConfigs)
+		TransferClusterHostStatesHandler(oc, nc)
```
wangfakang, line 204: `NewSimpleHostHandler` will be called in `NewSimpleHostHandler` , so it is not necessary to call `TransferClusterHostStatesHandler`.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -330,6 +346,77 @@ func newEdfLoadBalancerLoadBalancer(hosts types.HostSet, unWeightChoose func(typ
...
+
+//slowStartDurationFactorFuncWithNowFunc with nowFunc parameter for testing
+func slowStartDurationFactorFuncWithNowFunc(info types.ClusterInfo, host types.Host, nowFunc func() time.Time) float64 {
+	slowStart := info.SlowStart()
+
+	if slowStart.SlowStartWindow <= 0 {
+		return 1.0
+	}
+
+	duration := nowFunc().Sub(host.StartTime())
```
wangfakang, line 361: If the host's startime always uses the old, the host will only once a slow start.

jizhuozhi: No better solution, I will add comment for that

pkg/upstream/cluster/cluster_manager.go
jizhuozhi: `NewSimpleHostHandler(nc, hostConfigs)` just is `nc` but not transfer states from `oc` to `nc`

pkg/filter/stream/ipaccess/stream_filter_test.go
```diff
@@ -265,6 +266,14 @@ func (m *MockHost) SupportTLS() bool {
 	return false
 }
 
+func (m *MockHost) StartTime() time.Time {
```
alpha-baby, line 269: should use tool auto generate code, look this package `readme.md`

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -330,6 +346,72 @@ func newEdfLoadBalancerLoadBalancer(hosts types.HostSet, unWeightChoose func(typ
...
+
+	slowStart := info.SlowStart()
+
+	mode := slowStart.Mode
+	if mode == "" {
+		return hostWeightFunc
+	}
+
+	factorFunc := slowStartFuncFactories[mode]
+	if factorFunc == nil {
```
alpha-baby, line 385: add WARN log

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -319,7 +334,8 @@ func (lb *EdfLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) int {
 	return lb.hosts.Size()
 }
 
-func newEdfLoadBalancerLoadBalancer(hosts types.HostSet, unWeightChoose func(types.LoadBalancerContext) types.Host, hostWeightFunc func(host WeightItem) float64) *EdfLoadBalancer {
+func newEdfLoadBalancer(info types.ClusterInfo, hosts types.HostSet, unWeightChoose func(types.LoadBalancerContext) types.Host, hostWeightFunc func(host WeightItem) float64) *EdfLoadBalancer {
```
alpha-baby, line 337: unweightChooseHostFunc 
如果启动了慢启动 那么要避免走 unWeightChoose

pkg/filter/stream/ipaccess/stream_filter_test.go
jizhuozhi: > should use tool auto generate code, look this package `readme.md`

Sorry, I'm not found related codegen tools :(

alpha-baby: https://github.com/mosn/mosn/blob/master/pkg/mock/readme.md

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: This doesn't seem to be a problem by keeping the same logic as envoy

jizhuozhi: It seems only while failed to pick a healthy host will use unweighted choosing
https://github.com/mosn/mosn/blob/49542ae13495d539445aeb5825f2a3fe71d1bc25/pkg/upstream/cluster/loadbalancer.go#L325-L326

alpha-baby: here https://github.com/mosn/mosn/blob/49542ae13495d539445aeb5825f2a3fe71d1bc25/pkg/upstream/cluster/loadbalancer.go#L416

jizhuozhi: fixed by checking slow-start mode

jizhuozhi:
Hello, @wangfakang and @alpha-baby . Please take the time to see if there are other points that need to be modified :)

alpha-baby:
LGTM

wangfakang:
> Hello, @alpha-baby and @wangfakang . All conversations are resolved, PLAT

@jizhuozhi could you add a example for slow start. Here is a examples doc: https://github.com/mosn/mosn/tree/master/examples.  Thanks.

pkg/upstream/cluster/host.go
```diff
@@ -57,6 +58,7 @@ func NewSimpleHost(config v2.Host, clusterInfo types.ClusterInfo) types.Host {
 		tlsDisable:    config.TLSDisable,
 		weight:        config.Weight,
 		healthFlags:   GetHealthFlagPointer(config.Address),
+		startTime:     time.Now(),
```
wangfakang, line 61: Initialize only when slow start is enabled.

pkg/upstream/cluster/cluster_manager.go
wangfakang, line 204: trans it only when slow start is enabled. Same elsewhere.

pkg/upstream/cluster/host.go
jizhuozhi: I will deal with this later.

pkg/upstream/cluster/cluster_manager.go
jizhuozhi: I will deal with this later.

jizhuozhi: Done

pkg/config/v2/upstream.go
```diff
@@ -121,6 +123,13 @@ type DnsResolverConfig struct {
 	Attempts int      `json:"attempts,omitempty"`
 }
 
+type SlowStartConfig struct {
+	Mode              string              `json:"mode,omitempty"`
+	SlowStartDuration *api.DurationConfig `json:"slow_start_duration,omitempty"`
+	Aggression        float64             `json:"aggression,omitempty"`
+	MinWeightPercent  *Percent            `json:"min_weight_percent,omitempty"`
```
wangfakang, line 130: Why not use `float64` type directly?

pkg/types/upstream.go
```diff
@@ -23,7 +23,7 @@ import (
 	"sort"
 	"time"
 
-	metrics "github.com/rcrowley/go-metrics"
+	"github.com/rcrowley/go-metrics"
```
wangfakang, line 26: why delete metrics name?

pkg/upstream/cluster/host.go
```diff
@@ -35,28 +35,30 @@ import (
...
-		weight:        config.Weight,
-		healthFlags:   GetHealthFlagPointer(config.Address),
+		hostname:                config.Hostname,
+		addressString:           config.Address,
+		stats:                   newHostStats(clusterInfo.Name(), config.Address),
+		metaData:                config.MetaData,
+		tlsDisable:              config.TLSDisable,
+		weight:                  config.Weight,
+		healthFlags:             GetHealthFlagPointer(config.Address),
+		lastHealthCheckPassTime: time.Now(),
```
wangfakang, line 61: It's unnecessary to initialize the `lastHealthCheckPassTime`? Only need to assign a value when the health check changes.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -320,20 +336,97 @@ func (lb *EdfLoadBalancer) HostNum(metadata api.MetadataMatchCriteria) int {
...
+
+// slowStartDurationFactorFuncWithNowFunc with nowFunc parameter for testing
+func slowStartDurationFactorFuncWithNowFunc(info types.ClusterInfo, host types.Host, nowFunc func() time.Time) float64 {
+	slowStart := info.SlowStart()
+
+	if slowStart.SlowStartDuration <= 0 {
+		return 1.0
+	}
+
+	duration := nowFunc().Sub(host.LastHealthCheckPassTime())
```
wangfakang, line 363: if the value is true of `lastHealthCheckPassTime.IsZero()`, we should return `1.0` before `Sub` operation.

pkg/config/v2/upstream.go
jizhuozhi: As same as envoy `Percent`, I will fix it

pkg/types/upstream.go
jizhuozhi: IDE auto optimized

pkg/upstream/cluster/host.go
jizhuozhi: Compatible with XDS case. If the host is just discovered, then we need to discover its time as the basis for the slow start weight change. If the last health check pass time of discovered host is zero, we need an other field to record the time when discovered.

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: As above.

jizhuozhi: As above, at most situations, `lastHealthCheckPassTime` should not be zero, so it's a unnecessary checking.

pkg/upstream/cluster/host.go
wangfakang: This will cause all hosts to trigger the slow start strategy at the first time, so it could only take effect if the health flag changes before avoid the issue. Do you think Its ok?

jizhuozhi: I agree with your opinion. 

I just thought about it, maybe the health check pass time of the newly added host can be set in `transferHostSetStates`. 

This will only take effect when a new host is dynamically discovered through xds.

jizhuozhi: Fixed :)

jizhuozhi:
Hello, @wangfakang . All conversations are resolved. PLAT :)

wangfakang (approved):

alpha-baby (approved):
LGTM

Issue #2195, opened 2022-11-30
Support Setting Cluster Idle Timeout = 0


antJack:
Currently we only set upstream connection's idle timeout in case of cluster.idle_timeout > 0:
<img width="1012" alt="image" src="https://user-images.githubusercontent.com/52443884/204747611-b2432950-8bf3-49e9-99bb-d9291052b3af.png">

this issue requests the ability to set an idle_timeout = 0, to indicate the indefinite idle timeout for upstream connection timeout

PR #2197, opened 2022-12-05
Allow Setting Cluster Idle Timeout to Zero to Indicate Never Timeout

antJack:
ref: https://github.com/mosn/mosn/issues/2195


taoyuanyuan (approved):
LGTM

doujiang24 (approved):

PR #2202, opened 2022-12-19
[Feature] pprof debug server support endpoint configure

dengqian:

Support to configure debug server endpoint when we don't want to export debug server to user when using HostNetwork


3062 (approved):
LGTM

nejisama (approved):

doujiang24 (approved):

PR #2221, opened 2023-01-12
[bugfix]: metrics exclusionLabels and exclusionKeys not working for static config

dengqian:

```InitializeMetrics``` is called after ```initClusterManager``` and ```initServer```
cluster and host stats will be initialized before statsMatcher.ExclusionLabels and statsMatcher.ExclusionKeys set


doujiang24 (approved):

taoyuanyuan (approved):
LGTM

PR #2226, opened 2023-02-03
Update example server and client config file in README

mimani68:

While users are attempting to up and running the MOSN project, document enrichment would be extremely crucial to controlling the slope of learning curve. De facto clear addressing of config files in educational documents could be open floor to aspiring developers having more engaged to the project and enjoy the capabilities.

### Solutions

In the `/examples/en_readme/http-sample` we can feel the gap of recommended config file in term of easy-mode using project. To achieve this purpose the address of **current** and **stable** config file of **server** and **client** added to documentation.


taoyuanyuan (approved):
LGTM

doujiang24 (approved):

Issue #2185, opened 2022-11-23
edfheap#edfEntryLess is instable


hhdssmys:
### Describe the bug

[code link](https://github.com/mosn/mosn/blob/master/pkg/upstream/cluster/edfheap.go#L32) 
```go
func edfEntryLess(a, b *edfEntry) bool {
	if a.deadline == b.deadline {
		return a.weight < b.weight
	}
	return a.deadline < b.deadline
}
```
If heap sorting is done by weight, then nodes with higher weights may not be selected when ddl is the same. We should sort by edfEntry.queuedTime. What do you think? @alpha-baby 

### Environment

- MOSN Version

alpha-baby:
good, welcome commit PR to fix this.

jizhuozhi:
Hello, @hhdssmys .

Low weight priority is deliberate. You can imagine that in a short time window, we want as many nodes as possible to be selected instead of always being knocked down on the same server. If high weight priority is always As I just said, there are all high-weight nodes in a short window.

But I think the deadline seems to be a valid comparison indicator, and I will consider the benefits it will bring

jizhuozhi:
Hello, @hhdssmys and @alpha-baby . 

Since $Deadline = Deadline_{prev} + \frac{1}{Weight}$, when $Deadline_a = Deadline_b$ then we prefer to give smaller weight more chance to make traffic more smooth. Using the entry time for comparison seems to hope to give more opportunities to elements selected earlier, but Deadline itself plays the role of a logical clock, so the role of queuedTime is not so clear.

Comparing with $Weight$ and  comparing with $Deadline_{prev}$ are mathematically equivalent.

$$ Deadline_a < Deadline_b$$

$$ => Deadline - \frac{1}{Weight_a} < Deadline -  \frac{1}{Weight_b} $$

$$ => - \frac{1}{Weight_a} < - \frac{1}{Weight_b} $$

$$ =>  \frac{1}{Weight_a} > \frac{1}{Weight_b} $$

$$ => Weight_a < Weight_b$$

The QueuedTIme here only serves to record the previous time, but in the future, I tend to use the last Deadline as the enqueue time to maintain logical consistency, and welcome to submit relevant PRs

alpha-baby:
https://github.com/envoyproxy/envoy/blob/c58b4eaab93cdc13a786c246dd75cdb26b7a7ce0/source/common/upstream/edf_scheduler.h#L104

[...]

PR #2229, opened 2023-02-06
feat: use logical clock to make edf scheduler more stable

jizhuozhi:

#2185 

### Solutions
Use a logical clock to hold the enqueued time, to make sure FIFO order when both deadline and weight of entries are same.

As the proof in the https://github.com/mosn/mosn/issues/2185#issuecomment-1326078414, when deadline of entries are same, if entry `A` is left ordered to `B`, the weight of `A` is must less or equal to `B`, and the enqueued time is in the same order. But when fix the structure of the heap, we could not make sure left subtree is absolutely left ordered to the right, because binary heap is unstable. So here's a paradox where A and B enter the queue at the same time (since synchronization has been done), in order to avoid this paradox, we use a strictly self-increasing logical clock as the queue time.

### UT result
We generate random hosts and schedule them, checking their distribution every cycle. 

Ideally we should check that the sequence is the same every time, but because of the accumulation of floating point errors, there is no way to guarantee this, so we only check that the host distribution is as expected in a limited period.

```go
func TestEdfSchedulerDistribution(t *testing.T) {
	var weights []uint32
	totalWeights := uint32(0)

	rnd := func(low, high int) int {
		return rand.Intn(high-low) + low
	}

	checkDistribution := func(seq []string) {
		dist := make(map[string]int)
		for _, s := range seq {
			dist[s]++
		}
		for i, w := range weights {
			d := dist[fmt.Sprintf("host-%d", i)]
			assert.Equal(t, uint32(d), w)
		}
	}

	for i := rnd(64, 512); i >= 0; i-- {
		w := uint32(rnd(int(v2.MinHostWeight), int(v2.MaxHostWeight)))
		weights = append(weights, w)
		totalWeights += w
	}

	scheduler := newEdfScheduler(len(weights))
	for i, w := range weights {
		scheduler.Add(&mockHost{name: fmt.Sprintf("host-%d", i), w: w}, float64(w))
	}

	for i := 0; i < 1024; i++ {
		seq := make([]string, 0)
		for i := uint32(0); i < totalWeights; i++ {
			h := scheduler.NextAndPush(func(item WeightItem) float64 {
				return float64(item.Weight())
			}).(*mockHost)
			seq = append(seq, h.name)
		}
		checkDistribution(seq)
	}
}
```

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -19,7 +19,6 @@ package cluster
 
 import (
 	"sync"
-	"time"
 
 	"mosn.io/mosn/pkg/config/v2"
 )
@@ -28,6 +27,7 @@ type edfScheduler struct {
 	lock        sync.Mutex
 	items       *edfHeap
 	currentTime float64
+	clock       int64
 }
 
 func newEdfScheduler(cap int) *edfScheduler {
@@ -41,7 +41,7 @@ type edfEntry struct {
 	deadline   float64
 	weight     float64
 	item       WeightItem
-	queuedTime time.Time
+	queuedTime int64
 }
 
 type WeightItem interface {
@@ -57,7 +57,7 @@ func (edf *edfScheduler) Add(item WeightItem, weight float64) {
 		deadline:   edf.currentTime + 1.0/weight,
 		weight:     weight,
 		item:       item,
-		queuedTime: time.Now(),
+		queuedTime: edf.tick(),
 	}
 	edf.items.Push(&entry)
 }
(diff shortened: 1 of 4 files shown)


alpha-baby (approved):
LGTM

jizhuozhi:
Hello, @alpha-baby and @doujiang24 . PLAT, thx :)

doujiang24 (approved):

alpha-baby (approved):
LGTM

PR #2234, opened 2023-02-15
Feature/bugfix goaway

hui-cha:
### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/network/listener.go
```diff
@@ -223,12 +224,27 @@ func (l *listener) readMsgEventLoop(lctx context.Context) {
...
+		}
+		return err
+	} else {
+		// If it is not upgrade, close the listener first to avoid new connection establishment.
+		// These newly established connections will not be processed by mosn before listener closed, and the requests on the connection will timeout.
+		//
+		// Close listener should not close existsing connections.
+		err := l.Close(lctx)
+
+		// Close the existing connections if it support graceful close.
```
wangfakang, line 244: If the `bindToPort` of the listener is false, whether it is necessary to call `OnShutdown` and it is safe?

pkg/types/network.go
```diff
@@ -125,7 +125,7 @@ type Listener interface {
 	GetListenerCallbacks() ListenerEventListener
 
 	// Shutdown stop accepting new connections and graceful stop the existing connections
-	Shutdown() error
+	Shutdown(lctx context.Context) error
```
wangfakang, line 128: need update the API description.

pkg/network/listener.go
```diff
@@ -223,12 +224,27 @@ func (l *listener) readMsgEventLoop(lctx context.Context) {
 }
 
 // Shutdown stop accepting new connections and graceful close the existing connections
-func (l *listener) Shutdown() error {
-	changed, err := l.stopAccept()
-	if changed {
+func (l *listener) Shutdown(lctx context.Context) error {
```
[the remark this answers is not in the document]
hui-cha: Added some comments to describe the behavior in hot upgrade and non-hot upgrade scenarios

pkg/network/listener.go
hui-cha: Added a bindToPort status check here to keep it consistent with the original logic

pkg/types/network.go
hui-cha: Added some comments here

wangfakang (approved):
LGTM. ping @doujiang24 could you finaly review it

doujiang24 (approved):

Issue #2227, opened 2023-02-06
[bug] rand func used in healthcheck is not safe for concurrent access


dengqian:
panic info: 
```
 [ERROR] [healthcheck.session] [upstream] [health check] [session checker] panic runtime error: index out of range [-1]
goroutine 149 [running]:
runtime/debug.Stack()
    /tmp/usr/local/go/src/runtime/debug/stack.go:24 +0x65
gitlab.alibaba-inc.com/apsara-edge/mosn/vendor/mosn.io/mosn/pkg/upstream/healthcheck.(*sessionChecker).Start.func1()
    /home/admin/22_20221221191031931_443226823_code/rpm_workspace/go/src/gitlab.alibaba-inc.com/apsara-edge/mosn/vendor/mosn.io/mosn/pkg/upstream/healthcheck/session_checker.go:69 +0x46
panic({0x1c59040, 0xc0003db230})
    /tmp/usr/local/go/src/runtime/panic.go:1038 +0x215
math/rand.(*rngSource).Uint64(...)
    /tmp/usr/local/go/src/math/rand/rng.go:249
math/rand.(*rngSource).Int63(0x6)
    /tmp/usr/local/go/src/math/rand/rng.go:234 +0x9f
math/rand.(*Rand).Int63(...)
    /tmp/usr/local/go/src/math/rand/rand.go:84
math/rand.(*Rand).Int63n(0xc000e573e0, 0x4c4b40)
    /tmp/usr/local/go/src/math/rand/rand.go:116 +0x64
gitlab.alibaba-inc.com/apsara-edge/mosn/vendor/mosn.io/mosn/pkg/upstream/healthcheck.(*healthChecker).getCheckInterval(...)
```
This is because rand.Int63 was considered  non-concurrency-safe:
```
func (rng *rngSource) Uint64() uint64 {
	rng.tap--
	if rng.tap < 0 {
		rng.tap += rngLen
	}

	rng.feed--
	if rng.feed < 0 {
		rng.feed += rngLen
	}

	x := rng.vec[rng.feed] + rng.vec[rng.tap] // panic point
	rng.vec[rng.feed] = x
	return uint64(x)
}
```
So I think we should new a ```rand.Rand``` struct for every health checker, instead of a pointer
```
type healthChecker struct {
	//
	sessionConfig       map[string]interface{}
	sessionFactory      types.HealthCheckSessionFactory
	checkers            map[string]*sessionChecker
	localProcessHealthy int64
[...]

PR #2240, opened 2023-02-22
bugfix: healthcheck panic 

dengqian:
https://github.com/mosn/mosn/issues/2227 is not completely fixed.
The health check tasks of different hosts in the same cluster will reuse the same "rander", so calling Rand() concurrently still occasionally cause panic

### DIFF
--- a/pkg/upstream/healthcheck/healthchecker.go
+++ b/pkg/upstream/healthcheck/healthchecker.go
@@ -240,8 +240,8 @@ func (hc *healthChecker) getCheckInterval() time.Duration {
 	interval := hc.intervalBase
 	if hc.intervalJitter > 0 {
 		hc.mux.Lock()
-		defer hc.mux.Unlock()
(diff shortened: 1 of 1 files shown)


3062 (approved):

doujiang24 (approved):

pkg/upstream/healthcheck/healthchecker.go
```diff
@@ -237,6 +239,8 @@ func (hc *healthChecker) runCallbacks(host types.Host, changed bool, isHealthy b
 func (hc *healthChecker) getCheckInterval() time.Duration {
 	interval := hc.intervalBase
 	if hc.intervalJitter > 0 {
+		hc.mux.Lock()
+		defer hc.mux.Unlock()
```
taoyuanyuan, line 243: It is recommended not to use defer here, otherwise TODO will also be locked.

doujiang24 (approved):

taoyuanyuan (approved):

Issue #2252, opened 2023-03-17
L7 protocol supports latency-based load balancing algorithm


jizhuozhi:
### The New Feature
For L7 application layer protocol forwarding, a load balancing algorithm based on response time is provided to avoid forwarding requests to hosts with degraded service quality

### Your scenes
In service governance, we focus on three indicators, Rate, Error, and Duration (an interesting abbreviation called RED).

In the L4 transport layer, Round Robin and Least Active Connection (sometimes also called Least Outstanding Request, LOR) are generally implemented. This is because of the limited number of indicators held in the L4 transport layer.

But in the protocol (especially HTTP/HTTP2) in the L7 application layer, based on the request response mode, we can get the latency (Duration) of service request execution. In a cloud environment, services can be deployed on different servers, data centers, regions, servers of different scales, different models, and even different instruction sets, and the deployment forms are also different: bare metal, virtual machines, containers, etc. Therefore, the quality of service provided by hosts in different locations varies. Even if it is the same host, due to reasons such as busy disk IO, high CPU usage, garbage collection, etc., the quality of service at different times is different.

Only using Round Robin and Least Active Connection cannot avoid forwarding requests to these servers with reduced quality of service (the health check still passes, so the registry will not remove it). In order to maximize the quality of service, I expect to implement a load balancing algorithm based on latency, which can sense the decline in service quality, comprehensively respond to factors such as latency (Duration) and success rate (Error) to select the optimal server node as much as possible and ensure that it will not Choose the worst node.

#### References:
Linkerd: https://linkerd.io/2016/03/16/beyond-round-robin-load-balancing-for-latency/
Ingress-nginx: https://github.com/kubernetes/ingress-nginx/pull/2229
Envoy: https://github.com/envoyproxy/envoy/issues/20907 (WIP)

### Environment

- MOSN Version

jizhuozhi:
As described in #2295 , requests will skew when fail-fast occered, and #2296 tries to fix the client/server errors recording problem.

Because different errors have different reference meanings for load balancing, the revised formula can be changed. 

Since the client error is caused by the wrong behavior of the user, it basically has no impact on the system, so we can simply ignore this part of the request, but the server error is different, because the server really has an error, so the load balancing must avoid giving it sends traffic. 

So the calculation formula has changed a little, where `ClientErrorBias` and `ServerErrorBias` are configurable

$$
\begin{align}
alpha &= SuccessCount / TotalCount \\
[...]

PR #2253, opened 2023-03-17
feat: PeakEWMA load balancer

jizhuozhi:

#2252 

### Solutions
Based on the host indicators (`HostStats()`) that have been collected so far, use the average duration (perhaps a better percentile) as the characteristic of the response time, and use the number of active requests for weighting to calculate the expected time to process all current requests, and use the success rate for weighting to avoid Requests are forwarded to a large number of faulty servers. In order to avoid starvation of servers that have never succeeded, the minimum success rate weight is limited to 0.1.

In small clusters (differentiated by choice), choose the best among all hosts, and avoid choosing the worst in large clusters by random sampling (See [The Power of Two Random Choices: A Survey of Techniques and Results](http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf)).

### Benchmark

```
goos: darwin
goarch: amd64
pkg: mosn.io/mosn/pkg/upstream/cluster
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkShortestResponseLoadBalancer_ChooseHost
BenchmarkShortestResponseLoadBalancer_ChooseHost-16    	 3072984	       388.4 ns/op
PASS
```

### Code Style
+ Show `Golint` result


jizhuozhi:
Hello, @doujiang24 and @alpha-baby , PLAT, thx :)

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +625,98 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
 }
+
+type shortestResponseLoadBalancer struct {
+	hosts  types.HostSet
+	rand   *rand.Rand
+	mutex  sync.Mutex
+	choice uint32
+}
+
+func newShortestResponseLoadBalancer(info types.ClusterInfo, hosts types.HostSet) types.LoadBalancer {
```
jizhuozhi, line 636: There is a check failure reported by reviewdog, which requires we return struct (return `shortestResponseLoadBalancer` instead of `types.LoadBalancer`), but `RegisterLBType` not accepts this function, so this lint cannot be fixed.

doujiang24:
oops, let's disable golinter in this PR, to validate faster.

jizhuozhi:
> oops, let's disable golinter in this PR, to validate faster.

I can't laugh or cry. 😂

Cyclomatic complexity is a magical thing, I mean it is difficult to judge subjectively how much is appropriate. But I think it is at least a quantifiable indicator of mental stress.

I have refactored the code to reduce the cyclomatic complexity. I think it can be retained and used as an indicator to evaluate the quality of the code, unless it brings too much noise, we can let go of the quality door

doujiang24 (approved):

doujiang24:
Oh, it's better to add some comments somewhere, the new lb depends on metrics, it won't work when metrics are disabled.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +625,95 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+	var candidates []types.Host
+	for i := 0; i < choice; i++ {
+		idx := i
+		if rnd {
+			idx = lb.rand.Intn(total)
+		}
+		temp := hs.Get(idx)
+
+		// Newly added host always has the shortest response
+		if temp.HostStats().UpstreamRequestTotal.Count() == 0 {
```
nejisama, line 678: Metrics can be disabled by config.
if host metrics are disabled , this load balancer cannot use.
maybe should check it ?

jizhuozhi:
Made a series of optimizations, the latest benchmark
```
goos: darwin
goarch: amd64
pkg: mosn.io/mosn/pkg/upstream/cluster
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkShortestResponseLoadBalancer_ChooseHost
BenchmarkShortestResponseLoadBalancer_ChooseHost-16    	 4770519	       258.3 ns/op
PASS
```

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: Yes, this load balancer must working with metrics. And I didn't find where could disable the metrics, is there any document?

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +625,95 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+
+	lb.mutex.Lock()
+	defer lb.mutex.Unlock()
+
+	choice := total
+	rnd := false
+	// Choose `choice` times and return the best one.
+	// See The Power of Two Random Choices: A Survey of Techniques and Results
+	//  http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
+	if choice > int(lb.choice) {
```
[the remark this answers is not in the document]
wangfakang: Such as offline communication has been resolved.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +625,95 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+	total := hs.Size()
+
+	if total == 0 {
+		return nil
+	}
+	if total == 1 {
+		return hs.Get(0)
+	}
+
+	lb.mutex.Lock()
```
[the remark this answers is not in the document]
wangfakang: Such as offline communication has been resolved.

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: Hello, @nejisama . I didn't find the config which to disable the metrics. The only way that could make an impact is `feedback` when shm is not configured.

https://github.com/mosn/mosn/blob/master/pkg/metrics/shm/configs.go#L20-L21

jizhuozhi: Added logic to check and downgrade to WRR if disabled

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +626,144 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+	return lb.hosts.Size()
+}
+
+func (lb *shortestResponseLoadBalancer) iterateChoose() types.Host {
+	var candidate types.Host
+	var candidateScore float64
+
+	lb.hosts.Range(func(temp types.Host) bool {
+		if !temp.Health() {
+			return false
```
muyuan0, line 714: It seems an unhealthy node in the middle will break the iteration

pkg/upstream/cluster/loadbalancer_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"context"
 	"math"
 	"math/rand"
+	"mosn.io/mosn/pkg/metrics"
```
muyuan0, line 24: Interesting, why this import is not sorted?

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +626,144 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+		}
+
+		tempScore := shortestResponseScore(temp)
+		if candidate == nil || (tempScore < candidateScore || (tempScore == candidateScore && (temp.Weight() >= candidate.Weight()))) {
```
muyuan0, line 718: Maybe we can ignore the Weight and just use `tempScore < candidateScore`? Using the weight may be misleading as unlike other balancers, the weight here is just a priority for nodes that have the same score.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +626,144 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+func (lb *shortestResponseLoadBalancer) randomChoose() types.Host {
+	total := lb.hosts.Size()
+
+	var candidate types.Host
+	var candidateScore float64
+
+	for i := 0; i < int(lb.choice); i++ {
+		lb.mutex.Lock()
+		idx := lb.rand.Intn(total)
+		lb.mutex.Unlock()
```
muyuan0, line 745: The standard P2C requires the chosen host to be unique:
> For every request the balancer needs to route, it picks two unique instances at random.

https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/daperture-load-balancer

Maybe you are using the trace-off from the envoy? Like https://github.com/envoyproxy/envoy/issues/14859#issuecomment-770662594.

If so, let's add a comment for it.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +626,144 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+
+		temp := lb.hosts.Get(idx)
+		if !temp.Health() {
+			continue
```
muyuan0, line 749: Is it possible that every try hits the unhealthy host? In iterateChoose, the unhealthy hosts will be skipped. But in randomChoose, the unhealthy host will occupy the position of chosen hosts.

muyuan0, line 749: Is it possible that every try hits the unhealthy host? In iterateChoose, the unhealthy hosts will be skipped. But in randomChoose, the unhealthy host will occupy the position of chosen hosts.

muyuan0: Sorry, the iterateChoose will also return a nil value if every host in it is unhealthy. The randomChoose just has a higher chance to return a nil value.

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: This is deliberate, since we only use random selection when we guarantee a minimum of 3 elements.

In fact, it is very easy to ensure a unique random selection. Select k elements from N elements. In the i-th selection, we only need to use the following method to ensure a unique random selection

$$
P(i) = Rnd(P(i - 1), N - (k - i) + 1)
$$

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: When the user configures the weight, it must be intentional, and the weight should compound the distribution ratio as much as possible. Maybe I should use it as a weighting factor instead of judging when the scores are the same

pkg/upstream/cluster/loadbalancer.go
muyuan0: > This is deliberate, since we only use random selection when we guarantee a minimum of 3 elements.

I am sorry that I can't get you.
Even if the total is larger than choices, the same element may be chosen repeatedly if we choose it randomly.

pkg/upstream/cluster/loadbalancer.go
muyuan0: Weight is not so useful in shortestResponse or other similar response time driven balancers (like EWMA).
These balancers are considered self-balancing without giving a particular weight. For example, different sizes of nodes can have the number of requests match their sizes by making the response time equal.
Another issue is that P2C requires each host to have equal weight.

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: All right, I have guaranteed that the random selection will not have duplicate elements in newest commit

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: You are right, but most implemention of Least Outstanding Request (or Least Active Request) which using P2C alse supports weighted. I don't think it's a problem, if user doesn't want to make it weighted, just keeps it to 1.

jizhuozhi: > Weight is not so useful in shortestResponse or other similar response time driven balancers (like EWMA).

In fact, I am currently implementing the new branch (https://github.com/jizhuozhi/mosn/tree/ewma_balancer) at the fork of this branch to implement the shortest response time load balancing algorithm based on EWMA, and verify its effectiveness through repeated chaos tests.

In the latest fork I'm trying to incorporate response success rates for smarter traffic distribution.

For example, in a mixed deployment scenario, 1C, 2C, 4C, etc. exist at the same time. The slot-based concurrency control algorithm will manage the remaining available cores. When the core 1C server is overloaded and fail-fast, the load balancing based on response time Algorithms may mistakenly think this is best instead of considering servers with 2C or 4C (since they have a lot of active connections). But if we factor the response success rate into the calculation, we can avoid this from happening, because it knows that it is due to fast failure rather than it is really fast. 

Taking weight as a factor is also out of this consideration. Although we are randomly selecting hosts because of P2C, from a global perspective, the distribution ratio is in line with the weight ratio, which can also avoid this situation from happening.

muyuan0: > You are right, but most implemention of Least Outstanding Request (or Least Active Request) which using P2C alse supports weighted.

Could you provide some examples? AFAIK, envoy supports both P2C and weight, but it only uses P2C for unweight situation:
https://github.com/envoyproxy/envoy/blob/db6435ac397b6bb43a212ad83bfdfd425de5ff93/source/common/upstream/load_balancer_impl.cc#L1054
And ingress-nginx supports EWMA via P2C, but it doesn't support using weight in it.

The problem about using P2C with weight is that the P2C algorithm requires each node to have the same weight. For example, considering we have 100 nodes, one of them has 100 weight, and the rest have 1 weight. By using P2C, the 100 weight node only has almost 2% to be chosen to compare. So even if it is chosen each time after comparison, only 2% of requests will be forwarded to the node, which is much less than the ratio of weight (100 / (100 + 99)).

Anyway, this is beyond the topic of this PR.

muyuan0: > You are right, but most implemention of Least Outstanding Request (or Least Active Request) which using P2C alse supports weighted. I don't think it's a problem, if user doesn't want to make it weighted, just keeps it to 1.

BTW, we are discussing the weight in shortest response loadbalancer. Whether an implementation of Least Outstanding Request (or Least Active Request) uses weight or not is not so relative to the topic - as the Least Outstanding Request (or Least Active Request) is another loadbalancer.

muyuan0 (dismissed):

pkg/upstream/cluster/loadbalancer.go
jizhuozhi: Hello, @muyuan0 . There is an example that NGINX supports weighted connection when using P2C (named Random2 in NGINX). 
See: https://github.com/nginx/nginx/blob/master/src/http/modules/ngx_http_upstream_random_module.c#L365-L373

jizhuozhi: If you're interested, we can discuss it in a new PR (#2274 ) where more optimizations are done over the current implementation. And I am doing chaos testing in various scenarios, and I will attach the test report to the PR later, welcome to discuss and make suggestions :)

muyuan0: > Hello, @muyuan0 . There is an example that NGINX supports weighted connection when using P2C (named Random2 in NGINX). See: [nginx/nginx@`master`/src/http/modules/ngx_http_upstream_random_module.c#L365-L373](https://github.com/nginx/nginx/blob/master/src/http/modules/ngx_http_upstream_random_module.c?rgh-link-date=2023-03-30T14%3A43%3A00Z#L365-L373)

I am afraid you make a mistake here:
Nginx doesn't implement the standard P2C algorithm, at least it doesn't implement the P2C as this PR does.
Here it uses O(logn) to find the random node:
https://github.com/nginx/nginx/blob/dfe70f74a3558f05142fb552cea239add123d414/src/http/modules/ngx_http_upstream_random_module.c?rgh-link-date=2023-03-30T14%3A43%3A00Z#L407
It can't use O(1) to find the random node, otherwise, it will require O(n*weight) memory footprint.
However, this PR uses O(1) algorithm to select the node.

muyuan0: Another issue, as I pointed out in https://github.com/mosn/mosn/pull/2253#discussion_r1153315997, 
> BTW, we are discussing the weight in shortest response loadbalancer. Whether an implementation of Least Outstanding Request (or Least Active Request) uses weight or not is not so relative to the topic - as the Least Outstanding Request (or Least Active Request) is another loadbalancer.

you are talking about the Least Connections balancer. The link to the Nginx code is nothing about the shortest response balancer.

jizhuozhi: What I want to explain is that both Least Request and Shortest Response measure the processing capacity of the server to select the server with the fastest processing request, so their optimization ideas are interoperable.

The Least Request is based on the fact that water is injected into multiple leaking buckets at the same rate, and they are observed at the same time at a certain moment. The probability of the remaining water in the bucket is small is that the flow rate is fast. But when the outflow rate is faster than the injection rate, so that every observation is almost empty, then the number of connections at this time will not be able to support the decision. Although it is observed that the bucket is almost empty, the difference in the outflow rate still exists. 

On this basis, Shortest Response tries to collect samples at the end of the request and form a characteristic value accumulated by all historical requests. Although the characteristic value can fully reflect the server's processing capacity, we still need to avoid the last straw that crushes the camel's camel. This is the same as many EWMA algorithms use the number of connections for weighting (including linkerd and this implementation), or the response time is the thing called the weight of the number of connections.

As for nginx, yes, nginx does not implement standard P2C, it uses a weighted random algorithm. In fact, the concept of weight is always emphasized in the implementation of NGINX (just like RR is actually SWRR), it always hopes Give users a chance to control traffic, as I think so too. This is very common in actual deployment scenarios. Servers of different specifications are mixed in the same cluster and provide external services at the same time. There are many reasons for this. What I have experienced is that cloud service providers do not have enough resources in a certain availability zone Instances of the same specifications, then we can only purchase instances of different specifications and adjust the distribution ratio of traffic through weights

jizhuozhi: As for those users who do not want to be affected by the weight, it is very simple, just configure all nodes to have the same weight or not (the default is 1). 

Since the user configures the weight, there must be an intention to distribute it according to the weight. As a load balancing algorithm, it should satisfy this distribution method as much as possible. This is my idea.

muyuan0: > As for those users who do not want to be affected by the weight, it is very simple, just configure all nodes to have the same weight or not (the default is 1).

It is not as simple as it looks. We need to change the code according to the support of weight.
As I mentioned in https://github.com/mosn/mosn/pull/2253#discussion_r1153454010, if weight is needed to support, we can't have O(1) random choosing. I would recommend implementing P2C only for unweight branch, so in most case (the same weight is used) there is no need to use a complex O(logn) lookup.

```
	// If `total` is less than or equal to `choice`, we can iterate over all elements directly.
	if total <= int(lb.choice) {
		candidate = lb.iterateChoose()
	} else { // should also consider the weight here
		candidate = lb.randomChoose()
```

jizhuozhi: The weight factor is finally retained, but it is not used to distribute traffic according to the weight, but as a trade-off of processing capacity, just like the weighted least active request, it is an estimate of the processing time of active requests. In other words, called the number of active requests per unit resource

jizhuozhi:
Hello, @muyuan0 . According to https://github.com/mosn/mosn/pull/2274#discussion_r1154230917, it's renamed to `PeakEWMA` and the implementation is similar to Linkerd. PLAT 😺

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +626,171 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+		return nil
+	}
+
+	var candidate types.Host
+
+	// If `total` is less than or equal to `choice`, we can iterate over all elements directly.
+	if total <= int(lb.choice) {
+		candidate = lb.iterateChoose()
+	} else {
+		candidate = lb.randomChoose()
```
muyuan0, line 692: The random choose assumes every host has the same weight. Maybe we can add a check and use it only for unweight branch?

jizhuozhi: I may have misunderstood you.

In fact, neither random nor iterative, they do not distribute traffic according to weight. Iterative selection is just an optimization based on the fact that random selection is based on the fact that if n+k elements (k>0) are selected among n elements, then at least one element will be repeatedly selected (pigeon's nest principle), and for uniqueness and To select elements without repetition, it is to traverse all elements. So iterative selection is just a random selection of an optimization branch

muyuan0: So the weight here is not the same as the weight in the WRR? There will likely be an issue: the traffic distributions are different between random and iterative when the hosts have different weights. Each host is selected equally in random choose, so a large weight may not take effect as the probability of being selected is different from the weight. Maybe you can write a script with very imbalanced weights (like 1:10) to see if it is right or not.

jizhuozhi:
Hello, @muyuan0 .

After careful consideration, I decided to use the same implementation style as other LoadBalancers, use EdfLoadBalancer for weighted load balancing, and use unweighted branches when the weights are the same. This may not have optimal efficiency, but it can reduce the complexity of subsequent changes and maintenance.

In addition, due to the high cost of achieving random unique selection (the partition selection method will have skew when the score are the same), so still only using `rand` to ensure that the chosen is as unique as possible

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +625,146 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+//
+//	http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf
+func (lb *peakEwmaLoadBalancer) randomChoose() types.Host {
+	total := lb.hosts.Size()
+
+	var candidate types.Host
+
+	for i := 0; i < int(lb.choice); i++ {
+		lb.mutex.Lock()
+		idx := lb.rand.Intn(total)
```
muyuan0, line 737: > In addition, due to the high cost of achieving random unique selection (the partition selection method will have skew when the score are the same), so still only using rand to ensure that the chosen is as unique as possible

Let's add a comment to tell the readers that we don't ensure uniqueness because of the cost.

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +625,146 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+
+func (lb *peakEwmaLoadBalancer) iterateChoose() types.Host {
+	total := lb.hosts.Size()
+
+	var candidate types.Host
+
+	// avoid always choosing the same host when the score is the same
+	lb.mutex.Lock()
+	idx := lb.rand.Intn(total)
+	lb.mutex.Unlock()
```
muyuan0, line 703: IMHO, it is not so worth adding additional random generation & lock operations in the hot path.
Because when a host is chosen, its UpstreamRequestActive count will be increased, and its score is hardly the same the next time.

jizhuozhi: > Because when a host is chosen, its UpstreamRequestActive count will be increased, and its score is hardly the same the next time.

There is a special case here, that is, if metrics are disabled (https://github.com/mosn/mosn/pull/2253#discussion_r1147359057), then if we start traversing from a fixed position, all requests will be skewed to the same host.

And, I don't think the cost of the lock here will be very high, in the iterative branch, it will only be executed once, while in the random version it will be executed at least `choice` times

muyuan0: What a pity! Could you add a comment that this is for the special case?

jizhuozhi: > What a pity! Could you add a comment that this is for the special case?

Completed 😺

muyuan0 (approved):

doujiang24 (commented):
Thanks, basically LGTM.

pkg/metrics/store_lazy.go
```diff
@@ -188,3 +195,43 @@ func (lh *lazyHistogram) Variance() float64 {
...
+type lazyEWMA struct {
+	once sync.Once
+	ctor func() gometrics.EWMA
+	ewma gometrics.EWMA
+}
+
+// NewLazyEWMA build lazyEWMA
+func NewLazyEWMA(ctor func() gometrics.EWMA) (gometrics.EWMA, error) {
+	if ctor == nil {
+		return nil, ErrLazyEWMACtorEmpty
```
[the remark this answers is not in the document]
jizhuozhi: Consistent with other metrics, maybe we could open an another issue to address it?

pkg/metrics/store_lazy.go
```diff
@@ -188,3 +195,43 @@ func (lh *lazyHistogram) Variance() float64 {
...
 }
+
+type lazyEWMA struct {
+	once sync.Once
+	ctor func() gometrics.EWMA
+	ewma gometrics.EWMA
+}
+
+// NewLazyEWMA build lazyEWMA
+func NewLazyEWMA(ctor func() gometrics.EWMA) (gometrics.EWMA, error) {
```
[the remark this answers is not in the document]
jizhuozhi: As above

pkg/metrics/store_lazy.go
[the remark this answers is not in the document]
doujiang24: okay, let's leave it then.

doujiang24 (dismissed):
LGTM.
Looking forward to these TODOs, yep, they could be done in other PRs.
1. an example for this new usage, similar to: https://github.com/mosn/mosn/tree/master/examples/cn_readme/http-sample
2. a blog post to introduce this new lb.

jizhuozhi:
> LGTM. Looking forward to these TODOs, yep, they could be done in other PRs.

I will start next week

pkg/mosn/mosn.go
```diff
@@ -171,6 +172,19 @@ func (m *Mosn) initializeMetrics() {
...
 		log.StartLogger.Infof("[mosn] [init metrics] create metrics sink: %v", cfg.Type)
 	}
+
+	// set ewma alpha
+	if config.EWMAConfig != nil {
+		switch {
+		case config.EWMAConfig.Alpha > 0 && config.EWMAConfig.Alpha < 1:
+			cluster.SetAlpha(config.EWMAConfig.Alpha)
+		case config.EWMAConfig.Target > 0 && config.EWMAConfig.Target < 1 && config.EWMAConfig.Duration != nil:
+			cluster.SetAlpha(ewma.Alpha(config.EWMAConfig.Alpha, config.EWMAConfig.Duration.Duration))
```
jizhuozhi, line 182: When I was performing Chaos Tests, I found that result was not matching expectation. 

I'm sorry, I found this is a mistake

doujiang24 (dismissed):

pkg/proxy/upstream.go
```diff
@@ -85,9 +85,13 @@ func (r *upstreamRequest) OnDestroyStream() {}
 
 func (r *upstreamRequest) endStream() {
 	upstreamResponseDurationNs := time.Now().Sub(r.startTime).Nanoseconds()
+
 	r.host.HostStats().UpstreamRequestDuration.Update(upstreamResponseDurationNs)
+	r.host.HostStats().UpstreamRequestDurationEWMA.Update(upstreamResponseDurationNs)
```
taoyuanyuan, line 90: I think `EWMA` is a lb extension that the core framework should not be aware of.

taoyuanyuan: my mistake，lgtm

doujiang24 (approved):

taoyuanyuan:
@jizhuozhi  pls resolve the conflicting files

taoyuanyuan (approved):
LGTM

doujiang24 (approved):

taoyuanyuan:
pls add a doc
https://github.com/mosn/mosn.io/blob/master/content/zh/docs/products/configuration-overview/clustermanager/index.md

jizhuozhi:
> pls add a doc https://github.com/mosn/mosn.io/blob/master/content/zh/docs/products/configuration-overview/clustermanager/index.md

Hello @taoyuanyuan, I will do this. But since #2286 makes `activeRequestBias` configurable and PeakEWMA is using `activeRequest`, I will support it in PeakEWMA

PR #2258, opened 2023-03-21
[bugfix] sort addrs in strict dns cluster

dengqian:
Sort the addressed in case the LB policy will not work as expected

### DIFF
--- a/pkg/upstream/cluster/strict_dns_cluster.go
+++ b/pkg/upstream/cluster/strict_dns_cluster.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"reflect"
 	"runtime/debug"
+	"sort"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


doujiang24:
> Sort the addressed in case the LB policy will not work as expected

@dengqian could you please describe the unexpected behavior in detail? otherwise, LGTM.

dengqian:
> 

In some scenarios, the number of backend addresses does not change but the order of hosts changes.  When use Maglev LB strategy, the same index may forwarding to different backends

dengqian:
> > Sort the addressed in case the LB policy will not work as expected
> 
> @dengqian could you please describe the unexpected behavior in detail? otherwise, LGTM.

In some scenarios, the number of backend addresses does not change but the order of hosts changes. When use Maglev LB strategy, the same index may forwarding to different backends

doujiang24 (approved):
maybe we should describe a bit in the commit message, i.e.
make the LB behavior stable even when the order of the hosts changed.

3062 (approved):
LGTM

taoyuanyuan (approved):
LGTM

Issue #2256, opened 2023-03-20
 Some linter rules are too strict


jizhuozhi:
### Your question

In PR https://github.com/mosn/mosn/pull/2253#discussion_r1141576567, there is a rule `ireturn` (accept interfaces, return concrete structs), which requires all constructors of `loadbalancer` to return concrete types instead of interfaces, but because all factory patterns only accept registration methods that return interfaces. 

So this rule cannot be fixed, should we consider letting go of this rule?

doujiang24:
yep, I think we can disable the `ireturn` golint.

doujiang24:
Could you please try to disable it? another PR should be fine.

3062:
The two functions newRandomLoadBalancer and newReqRoundRobinLoadBalancer don't report an error, just because the return is written differently?

jizhuozhi:
> The two functions newRandomLoadBalancer and newReqRoundRobinLoadBalancer don't report an error, just because the return is written differently?

It's reported, see
https://github.com/mosn/mosn/actions/runs/4446019460/jobs/7805831517?pr=2253#step:3:1841 and https://github.com/mosn/mosn/actions/runs/4446019460/jobs/7805831517?pr=2253#step:3:1847

3062:
> > The two functions newRandomLoadBalancer and newReqRoundRobinLoadBalancer don't report an error, just because the return is written differently?
> 
> Hello, @3062 . It's reported, see https://github.com/mosn/mosn/actions/runs/4446019460/jobs/7805831517?pr=2253#step:3:1841 and https://github.com/mosn/mosn/actions/runs/4446019460/jobs/7805831517?pr=2253#step:3:1847
OK, maybe I asked a stupid question

jizhuozhi:
> Could you please try to disable it? another PR should be fine.

@doujiang24 Completed, see #2259 :)

PR #2259, opened 2023-03-21
Disable over strict linters in workflow

jizhuozhi:

#2256 

### Solutions
Disable `ireturn` in workflow file


doujiang24 (dismissed):

doujiang24:
Could you please also check this message? there might be other lints that need to be disabled.
https://github.com/mosn/mosn/actions/runs/4477912086/jobs/7870029509#step:3:94

The golint CI show happy may be due to some other reason, I mean it actually is unhappy.

antJack (approved):
lgtm

Issue #2260, opened 2023-03-21
Make metrics sampling configurable


jizhuozhi:
### The New Feature
A configurable way to select the sample implement when construct `Histogram`.

### Your scenes
`gometrics` package supports 4 sample implements: `NilSample`, `SampleSnapshot`, `UniformSample`, `ExpDecaySample`.

`NilSample` and `SampleSnapshot` are utils and not for `Histogram`. `UniformSample` using random sample and limited slots to hold the samples. `ExpDecaySample` also using random sample but keeping evicting older values (by priority queue with decaying random weight). 

`UniformSample` is simple but older sample which has weight as same as newest, but it can no longer reflect the real situation of the system. `ExpDecaySample` is more complex because the `alpha` (decay factor) is hard to decision, but in some intelli controller (eg. LoadBalancer), it's useful because it reflect the nearly real situation of the system.

### Your advice
It is currently difficult to bind the configuration to a certain metric, so it is possible to use a global variable and modify it through the configuration file.

### Environment

- MOSN Version

PR #2261, opened 2023-03-21
feat: make metrics sampling configurable

jizhuozhi:

#2260 

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/metrics/store.go
+++ b/pkg/metrics/store.go
@@ -25,6 +25,7 @@ import (
 	"sort"
 
 	gometrics "github.com/rcrowley/go-metrics"
+	"mosn.io/mosn/pkg/log"
 	"mosn.io/mosn/pkg/metrics/shm"
 	"mosn.io/mosn/pkg/types"
 )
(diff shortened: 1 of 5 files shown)


alpha-baby (dismissed):
LGTM

pkg/metrics/store.go
```diff
@@ -184,10 +185,30 @@ func (s *metrics) Histogram(key string) gometrics.Histogram {
 		return gometrics.NilHistogram{}
 	}
 
-	construct := func() gometrics.Histogram {
-		// TODO: notice the histogram only keeps 100 values as we set
-		return s.registry.GetOrRegister(key, func() gometrics.Histogram { return gometrics.NewHistogram(gometrics.NewUniformSample(100)) }).(gometrics.Histogram)
+	var construct func() gometrics.Histogram
+	if sampleType == SampleUniform {
```
doujiang24, line 189: better use switch instead of `if`?

doujiang24 (commented):
otherwise, LGTM.

pkg/metrics/store.go
jizhuozhi: I will do it

doujiang24 (approved):

pkg/metrics/store.go
```diff
@@ -25,6 +25,7 @@ import (
 	"sort"
 
 	gometrics "github.com/rcrowley/go-metrics"
```
spacewander, line 27: Please keep a blank line between third-party module and the module in this repo

pkg/metrics/store_test.go
```diff
@@ -19,6 +19,7 @@ package metrics
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/assert"
```
spacewander, line 22: Please configure goimports to format code

pkg/metrics/store.go
```diff
@@ -184,10 +185,31 @@ func (s *metrics) Histogram(key string) gometrics.Histogram {
...
+	case SampleExpDecay:
+		construct = func() gometrics.Histogram {
+			return s.registry.GetOrRegister(key, func() gometrics.Histogram {
+				return gometrics.NewHistogram(gometrics.NewExpDecaySample(sampleSize, expDecayAlpha))
+			}).(gometrics.Histogram)
+		}
+	default:
+		if log.DefaultLogger.GetLogLevel() >= log.WARN {
+			log.DefaultLogger.Warnf("[metrics] Unknown sample type %s, will use UNIFORM as default", sampleType)
+		}
```
spacewander, line 205: It would be better to do the fallback in SetSampleType, instead of when we need to use it?

pkg/metrics/store_test.go
jizhuozhi: I'm sorry, I will do this

pkg/metrics/store.go
jizhuozhi: Hello, @spacewander . 

I agree with you, but I don't like doing too much in the setters. And I realized that it shouldn't be limited to only these two types. In the case of package references (such as layotto), users should be able to expand more sampling logic. 

So I did some refactoring to support user registration of custom sampling types 😺

spacewander: If the user configures an incorrect value in the configuration (probably via SetSampleType), we should reject it at the beginning, instead of throwing a warning when creating each upstream/downstream stat.

jizhuozhi: I understand what you mean, I will check if the type exists in the map in the setter

spacewander (approved):

doujiang24 (approved):

Issue #2266, opened 2023-03-25
The time to establish a connection is lost in the request duration metric


jizhuozhi:
### Describe the bug

In the complete request chain, it includes establishing a connection (or acquiring a connection from the connection pool), writing request data to connection, reading response data from connection, closing the connection (or putting it back in the pool). No coverage for connection establishment duration has been found in the code so far.

The problem now is that when implementing intelligent load balancing and performing Chaos Testing (CT), it is found that when the quality of the basic network deteriorates (high latency or high packet loss rate), it is not reflected in the metrics, so the load balancing algorithm cannot perceive this change (resulting in incomplete functions ).

### Expected behavior

The time to establish a connection should also be reflected in the total request time, or have a separate metric reflecting the total time taken to establish a connection and complete a request. Because what we need to observe is the SLA of the service provided to the end user, not just how long the service took to process it.

### Actual behavior

When the underlying network quality deteriorates, we do not observe an impact on end-user quality of service.

### Steps to reproduce

Change the network quality of the host that provides the service (through the Chaos Test Agent or other means), and perform performance tests through the mosn proxy, and observe the change of `UpstreamRequestDuration` in mosn proxy.

### Minimal yet complete reproducer code (or GitHub URL to code)

Record `startTime` before acquiring stream

```diff
 pkg/proxy/upstream.go | 3 +--
 1 file changed, 1 insertion(+), 2 deletions(-)

diff --git a/pkg/proxy/upstream.go b/pkg/proxy/upstream.go
index 5dc34a526..c12998054 100644
--- a/pkg/proxy/upstream.go
+++ b/pkg/proxy/upstream.go
@@ -169,6 +169,7 @@ func (r *upstreamRequest) appendHeaders(endStream bool) {
                failReason   types.PoolFailureReason
        )

+       r.startTime = time.Now()
        if r.downStream.oneway {
                _, streamSender, failReason = r.connPool.NewStream(r.downStream.context, nil)
        } else {
@@ -234,8 +235,6 @@ func (r *upstreamRequest) OnReady(sender types.StreamSender) {

        r.requestSender = sender
        r.requestSender.GetStream().AddEventListener(r)
-       // start a upstream send
-       r.startTime = time.Now()

        if trace.IsEnabled() {
                span := trace.SpanFromContext(r.downStream.context)

```

### Environment

- MOSN Version

PR #2267, opened 2023-03-26
bugfix: request duration reflect the time to establish connection

jizhuozhi:

#2266 

### Solutions
Record request start time before getting connection.

### DIFF
--- a/pkg/metrics/upstream.go
+++ b/pkg/metrics/upstream.go
@@ -24,7 +24,7 @@ import (
 // UpstreamType represents upstream metrics type
 const UpstreamType = "upstream"
 
-//  key in cluster/host
+// key in cluster/host
 const (
 	UpstreamConnectionTotal                        = "connection_total"
 	UpstreamConnectionClose                        = "connection_close"
@@ -45,11 +45,13 @@ const (
 	UpstreamRequestPendingOverflow                 = "request_pending_overflow"
 	UpstreamRequestDuration                        = "request_duration_time"
 	UpstreamRequestDurationTotal                   = "request_duration_time_total"
+	UpstreamProxyDuration                          = "proxy_duration_time"
+	UpstreamProxyDurationTotal                     = "proxy_duration_time_total"
 	UpstreamResponseSuccess                        = "response_success"
 	UpstreamResponseFailed                         = "response_failed"
 )
 
-//  key in cluster
+// key in cluster
 const (
 	UpstreamRequestRetry         = "request_retry"
 	UpstreamRequestRetryOverflow = "request_retry_overflow"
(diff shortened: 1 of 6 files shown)


doujiang24:
Cool! I do think it's a problem. But, I'm not sure if it's a proper fix.
A bit worried about the backward compatibility.

nejisama:
The NewStream of ConnPool is established asynchronously, and a NewStream may establish up to 3 connections at a time.
When the next request is made to try NewStream, it may use the asynchronously established connection of the previous request. At this time, the time consumption is relatively low, but it does not mean that the corresponding target machine network is better than the previous one.

jizhuozhi:
Hello, @doujiang24 and @nejisama .

My intention is not to accurately represent the quality of service of the network, but to reflect the complete request processing time (that is, the time perceived by the end user).

In scenarios where pools are used or protocols that support multiplexing, only initializing the connection will bring additional overhead. After the connection is initialized, it should take the same time as the business request processing time. But when the quality of service of the network drops, the connection will never be added to the pool due to the connection timeout, so the connection will be continuously established. At this time, the delay will increase up to the connection timeout time, but it cannot be found in the metrics.

Or should a new metric be created to reflect this duration?

nejisama:
Maybe we should add a new metrics, and count it at real connect

jizhuozhi:
> Maybe we should add a new metrics, and count it at real connect

How about `UpstreamProxyDuration`? Because it's really proxied to upstream ~~

pkg/metrics/upstream.go
```diff
@@ -45,11 +45,13 @@ const (
 	UpstreamRequestPendingOverflow                 = "request_pending_overflow"
 	UpstreamRequestDuration                        = "request_duration_time"
 	UpstreamRequestDurationTotal                   = "request_duration_time_total"
+	UpstreamProxyDuration                          = "proxy_duration_time"
```
doujiang24, line 48: better add some comments.

doujiang24 (commented):
a small comment, otherwise, LGTM.

jizhuozhi:
The meaning of `UpstreamRequestDuration` has changed. In a word:
- `UpstreamRequestDuration`: The time from the initiation to the end of the request, including internally obtaining available connections and other processing

- `UpstreamProxyDuration`: Same meaning as the original `UpstreamRequestDuration`, only includes the processing time from writing the request to receiving the response

For Ops, the request-response interval should completely represent the complete time from the initiation to the end of the request, including obtaining the available connection and sending the request data until receiving the response from the server. But now `UpstreamRequestDuration` does not include the available connection time, which will bring costs for Ops: why there is such a big gap between the response time seen from the view of mosn and the view of end users.

To fill this gap, we revised the meaning of `UpstreamRequestDuration` and introduced a new `UpstreamProxyDuration` to replace the original `UpstreamRequestDuration`.

jizhuozhi:
cc @rayowang

pkg/metrics/upstream.go
spacewander, line 48: Could you add some comments to show the difference between request/proxy duration?

PR #2274, opened 2023-03-29
feat: PeakEWMA load balancer with response success rate

jizhuozhi:

#2252 

### Solutions
Forked from #2253. But two differents.

#### Duration with EWMA
The first is that using the two level implementation of an EWMA to calculate duration characteristics. The first level will be counted according to the second time interval to get the arithmetic mean, and then decay it through the exponential moving weighted average (EWMA).

#### Success Rate with EWMA
The second is that additional success rate metrics added. It's for fail-fast scenario.

For example, in a mixed deployment scenario, 1C, 2C, 4C, etc. exist at the same time. The slot-based concurrency control algorithm will manage the remaining available cores. When the core 1C server is overloaded and fail-fast, the load balancing based on response time Algorithms may mistakenly think this is best instead of considering servers with 2C or 4C (since they have a lot of active connections). But if we factor the response success rate into the calculation, we can avoid this from happening, because it knows that it is due to fast failure rather than it is really fast.

The success rate also using two-level EWMA to achieve the success rate of recession, avoiding the problem of low sensitivity of the arithmetic mean under very large samples.

### Benchmark

```
goos: linux
goarch: amd64
pkg: mosn.io/mosn/pkg/upstream/cluster
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkShortestResponseLoadBalancer_ChooseHost
BenchmarkShortestResponseLoadBalancer_ChooseHost-4   	 4357245	       287.4 ns/op
PASS
```

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/upstream/cluster/loadbalancer.go
+++ b/pkg/upstream/cluster/loadbalancer.go
@@ -743,16 +743,20 @@ func (lb *intelliLoadBalancer) randomChoose() types.Host {
 	var candidate types.Host
 	var candidateScore float64
 
-	lastIdx := 0
+	// ceil(total / choice), the number of segments for random selection
+	s := (total + int(lb.choice) - 1) / int(lb.choice)
 
-	for choice := int(lb.choice); choice > 0; choice-- {
-		n := total - lastIdx - choice
+	for i := 0; i < int(lb.choice); i++ {
+		begin := i * s
+		end := begin + s
+
+		if end > total {
+			end = total
+		}
 
 		lb.mutex.Lock()
-		idx := lb.rand.Intn(n) + 1 + lastIdx
+		idx := randRangeInt(lb.rand, begin, end)
 		lb.mutex.Unlock()
-
-		lastIdx = idx
 
 		temp := lb.hosts.Get(idx)
 		if !temp.Health() {
@@ -774,6 +778,10 @@ func (lb *intelliLoadBalancer) randomChoose() types.Host {
 	return candidate
 }
 
+func randRangeInt(rand *rand.Rand, start, end int) int {
+	return rand.Intn(end-start) + start
+}
+
 // All factors cannot be 0 to ensure that other factors can take effect normally
 const defaultFactorValue = 1e-6
 
@@ -793,7 +801,7 @@ func intelliScore(h types.Host) float64 {
 		// so calculate the number of active requests per unit of
 		// computing resources in the upstream through the weights,
 		// and then use `math.Log` to reduce impact.
-		activeRequests = math.Log((activeRequests + 1) / float64(h.Weight()))
+		activeRequests = math.Log(activeRequests/float64(h.Weight()) + 1)
 	}
 
 	duration := stats.UpstreamRequestDurationEWMA.Rate()
--- a/pkg/upstream/cluster/loadbalancer.go
+++ b/pkg/upstream/cluster/loadbalancer.go
@@ -798,5 +798,5 @@ func intelliScore(h types.Host) float64 {
 		responseSuccessRate = defaultFactorValue
 	}
 
-	return duration * float64(stats.UpstreamRequestActive.Count()) / float64(h.Weight()+1) / (responseSuccessRate * responseSuccessRate)
+	return duration * float64(stats.UpstreamRequestActive.Count()+1) / float64(h.Weight()) / (responseSuccessRate * responseSuccessRate)
 }
--- a/pkg/upstream/cluster/loadbalancer.go
+++ b/pkg/upstream/cluster/loadbalancer.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	gometrics "github.com/rcrowley/go-metrics"
 	"github.com/trainyao/go-maglev"
 
 	"mosn.io/api"
@@ -69,6 +70,7 @@ func init() {
 	RegisterLBType(types.Maglev, newMaglevLoadBalancer)
 	RegisterLBType(types.RequestRoundRobin, newReqRoundRobinLoadBalancer)
 	RegisterLBType(types.LeastActiveConnection, newleastActiveConnectionLoadBalancer)
+	RegisterLBType(types.ShortestResponse, newShortestResponseLoadBalancer)
 
 	RegisterSlowStartMode(types.ModeDuration, slowStartDurationFactorFunc)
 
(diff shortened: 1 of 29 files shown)


jizhuozhi:
It's normal to fail because of the dependency on https://github.com/mosn/api/pull/54. I will move it to draft status

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -624,3 +626,167 @@ func (lb *reqRoundRobinLoadBalancer) IsExistsHosts(metadata api.MetadataMatchCri
...
+		responseSuccessRate = 1e-6
+	}
+
+	duration := stats.UpstreamRequestDurationEWMA.Rate()
+	if duration == 0 {
+		duration = 1e-6
+	}
+
+	// All factors cannot be 0 to ensure that other factors can take effect normally
+	return duration * float64(stats.UpstreamRequestActive.Count()+1) / float64(h.Weight()) / responseSuccessRate
```
muyuan0, line 791: One question, why do we need to multiply the duration by the number of requests?
As the duration is the result of the number of requests. For a given node, a high number of requests usually mean a higher duration. Multiplying these two factors seems to amplify the difference. The ingress-nginx's ewma implementation only considers the duration.

Multiplying the number of requests has a side effect - now the shortest load balancer algorithm is not self-balancing. Considering two nodes (1C and 2C) have the same duration, they usually have different numbers of requests, for example, the 2C node may have a double number of requests than the 1C. Therefore, we need to configure a weight for the 2C node, otherwise they are not balanced. However, if we remove the number of requests, we don't need to introduce weights for different nodes.

jizhuozhi: In the real world, there is a singularity in the performance curve of the server with the number of requests processed at the same time (usually related to the number of CPU cores or other hardware resources). Before reaching this singularity, the performance is very good, but when it reaches this A sudden change in server performance occurs during the singularity.

To give a practical scenario, in a CPU-intensive service, before the number of CPU cores is less than the number of CPU cores, the service always performs well and has a uniform performance, but it will change when the number of concurrently processed requests exceeds the number of CPU cores. Either the frequent context switching causes the delay of all requests to increase, or it enters the queue for queuing, so as the queue gets longer and longer, the waiting time becomes longer and longer.

Therefore, although the performance of this server is very good, we still need to consider their load situation to avoid becoming the last straw of the compressed camel.

jizhuozhi: Linkerd also uses the EWMA algorithm weighted by the number of connections. This part of the logic is not in linkerd itself, but in tower-balance

See: https://github.com/linkerd/linkerd2/blob/cea37ee48d7cc93b5035598bf9b4da0db5a9e300/proxy/src/outbound.rs#L86-L89
and 
https://github.com/tower-rs/tower/blob/0ab48a2e6db55fabaa95370f304fbc0caf0616a6/tower-load/src/peak_ewma.rs#L175-L190

muyuan0: I can agree with some of your concerns.
However, the current implementation focuses more on the number of connections than the response duration.

There are two ways that the number of connections can affect the final result:
1. the larger number of connections, the longer duration
2. the score is multiplied by the number of connections

while the response duration only has a way to affect the final result.

> From another point of view, EWMA can also be considered as the weighting of the number of active requests, which is used to predict which server will reach the empty state faster (there is such an informal statement)

Yes, it is exactly what the current implementation does.

I would recommend renaming this loadbalancer to `LeastRequestWithEWMA`, so its name can reflect what it is. In fact, this loadbalancer will select a node with the least request instead of the shortest duration. For example,

Considering we have two nodes, node A has 2C, and node B has 1C. The number of active requests on A is 3, and 2 on B. As A has 2C, the estimated response duration of A is smaller than B (3/2 < 2/1).
For a loadbalancer which will always select the node with the shortest response, A will be chosen.
However, for the loadbalancer in this PR,
A has score 4.5, and B has score 4. B will be chosen because it has a lower score, even it has a higher estimated response duration.

Maybe a new name can solve the conflict and reach an agreement.

jizhuozhi: I see what you mean, I will try to implement a lock-free real-time EWMA, so that we can no longer use the number of connections and weights, but simply load balance based on response time.

jizhuozhi: I think I understand what you mean. Indeed, this is a problem especially when the difference in response time (std.) is orders of magnitude different from the response time itself, then changes in the number of connections will be more sensitive than the response time. But on the other hand, we can't just rely on the response time. As mentioned earlier, many times the increase in the response time is a jump. It is possible to use Round Robin without encountering a singularity, but because the EWMA will not be updated before the request returns. , then a large number of requests may flood into the optimal node at the same time, leading to the singularity.

I will still keep the number of connections factor, but I will try to reduce its impact, the logarithm may be a good means

muyuan0: > But on the other hand, we can't just rely on the response time.

Totally agree.

> I will still keep the number of connections factor

It is fine to keep the number of connections. But I think we can find a better name for this loadbalancer. There are many factors to select a node in it, the shortest response is just one of them. And this loadbalancer doesn't guarantee the node with the shortest response is chosen, even if the weights are equal. Unlike the least request loadbalancer, which will choose the node with the least active requests. I think this name is misleading and causes many conflicts. Maybe we can rename it, for instance, calling it EWMA or something else.

jizhuozhi: Hello, @muyuan0 . The name of the load balancing algorithm has been changed. 

And you are right, it has been verified that the weights cannot simply be factored into the calculation, and the cost of allocation according to the flow is very high (more than the benefit), so I try to use this as a hint to the number of requests: usually, more weights mean more computing resources, so calculate the number of active requests per unit of computing resources in the upstream through the weights, which is the same as the number of weighted active requests. 

Thank you very much for your advice and patience explanation of.

jizhuozhi:
Here is the formula for EWMA

$$
S_t = \alpha * i + (1 - \alpha) * S_{t-1}
$$

A constant called alpha appears in this formula. Many implementations will require the value of alpha to be configured, but there is no guidance on how to configure this value. 

In this PR, the specific calculation method is given: the alpha value required to decay from 1 to a sufficiently small value $\beta$ within a fixed time $t$. Let  $S_0 = 1$ then

$$
S_t = \alpha * 0 + (1 - \alpha) * S_{t-1} = (1 - \alpha) * S_{t-1} = (1 - \alpha) ^ t = \beta
$$

It equals to  $alpha = 1 - \beta^{-t}$. So users no longer need to pay attention to how to configure $\alpha$, only need to know the expected $\beta$ and $t$

jizhuozhi:
The algorithm has done a lot more work, and now we can specify our purpose of this load balancer (in order of priority)

1. Most importantly, the load balancer should choose different hosts, and should avoid overloading them, and avoid starvation of a server at the same time

2. We should avoid hosts with a high failure rate as much as possible.

3. Under the premise of ensuring that the request can be successfully processed, we should choose the one with the shortest processing time as much as possible

4. Regardless of the response time or the success rate, as long as there is a change, we should sense and make adjustments in a short time

This is also the Rate, Error and Duration (RED) corresponding to the quality of service.

Issue #2295, opened 2023-04-20
Support recording client errors and server errors on the L7 protocol


jizhuozhi:
### The New Feature

I want the stats of hosts supprt recording client errors (4xx in HTTP) and server errors (5xx in HTTP) on the L7 protocol.

### Your scenes

Currently the L7 response simply records success and failure, but there are different types of errors in L7 protocol (the HTTP protocol is well defined).
https://github.com/mosn/mosn/blob/92f428f1255a9ae0622beaf4fb11812ad5d93d23/pkg/types/upstream.go#L281-L282

I am currently implementing a latency based load balancing algorithm #2252 , but I'm facing a sticky problem now.

Latency-based load balancing algorithms are all based on the premise that the cost of processing requests is the same. But as we all know, user input is untrustworthy, and all backend services will verify user (or other service) input, and once the verification fails, the request will not be processed and the result will be returned quickly, which will lead to The load balancing algorithm gets a wrong view: this request is processed very quickly.

$$
\begin{align}
Avg &= \sum_{i=1}^{n}Duration_i / TotalCount \\
       &= (\sum_{i=1}^{SuccessCount}SuccessDuration_i + \sum_{i=1}^{FailedCount}FailedDuration_i) / TotalCount \\
        &\approx \sum_{i=1}^{SuccessCount}SuccessDuration_i / TotalCount \\
        &\ll \sum_{i=1}^{SuccessCount}SuccessDuration_i / SuccessCount
\end{align}
$$

We need a mathematical model to fix this wrong view. One of the easiest ways to do this is to fix all response times to the same value as the `SuccessDuration`

$$
\begin{align}
FixedAvg &= Avg / \alpha \\
\sum_{i=1}^{SuccessCount}SuccessDuration_i / SuccessCount &= \sum_{i=1}^{SuccessCount}SuccessDuration_i / TotalCount / \alpha \\
SuccessCount &= \alpha TotalCount \\
\alpha &= SuccessCount / TotalCount \\
\alpha &= SuccessRate
\end{align}
$$

But now client errors are included in $SuccessCount$.

### Your advice

It seems that we only need to record the failed requests count include client errors to solve the problem? 

In fact, different errors have different bias for the final score. The client error is only due to the abnormal behavior of the user. In fact, the host has no performance problems, but the server error is different. It means that there is a real problem with the host. If the request is sent to it again, it will seriously affect the quality of service.

[...]

PR #2296, opened 2023-04-20
feat: support recording HTTP status codes

jizhuozhi:

#2295

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/proxy/downstream.go
+++ b/pkg/proxy/downstream.go
@@ -1221,20 +1221,17 @@ func (s *downStream) handleUpstreamStatusCode() {
 	if s.upstreamRequest != nil && s.upstreamRequest.host != nil {
 		responseCode := s.requestInfo.ResponseCode()
 
-		// 4xx && 5xx
-		if responseCode >= http.BadRequest {
+		switch {
+		case responseCode >= http.BadRequest && responseCode < http.InternalServerError:
+			s.upstreamRequest.host.HostStats().UpstreamResponseClientError.Inc(1)
+			s.upstreamRequest.host.ClusterInfo().Stats().UpstreamResponseClientError.Inc(1)
+		case responseCode >= http.InternalServerError:
+			// TODO(jizhuozhi): `UpstreamResponseFailed` include client errors or configurable?
 			s.upstreamRequest.host.HostStats().UpstreamResponseFailed.Inc(1)
 			s.upstreamRequest.host.ClusterInfo().Stats().UpstreamResponseFailed.Inc(1)
-
-			if responseCode < http.InternalServerError {
-				s.upstreamRequest.host.HostStats().UpstreamResponseClientError.Inc(1)
-				s.upstreamRequest.host.ClusterInfo().Stats().UpstreamResponseClientError.Inc(1)
-			} else {
-				s.upstreamRequest.host.HostStats().UpstreamResponseServerError.Inc(1)
-				s.upstreamRequest.host.ClusterInfo().Stats().UpstreamResponseServerError.Inc(1)
-			}
-
-		} else {
+			s.upstreamRequest.host.HostStats().UpstreamResponseServerError.Inc(1)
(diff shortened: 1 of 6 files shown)


jizhuozhi:
Hello, @doujiang24 , @taoyuanyuan and @alpha-baby , PLAT, THX~

pkg/proxy/downstream.go
```diff
@@ -1217,11 +1218,22 @@ func (s *downStream) onUpstreamHeaders(endStream bool) {
 }
 
 func (s *downStream) handleUpstreamStatusCode() {
-	// todo: support config?
 	if s.upstreamRequest != nil && s.upstreamRequest.host != nil {
-		if s.requestInfo.ResponseCode() >= http.InternalServerError {
+		responseCode := s.requestInfo.ResponseCode()
+
+		// 4xx && 5xx
+		if responseCode >= http.BadRequest {
```
doujiang24, line 1225: this could be a break change, I'm not sure if it is a proper choice.

we'd better add some notes for this change, i.e. the description comment of this PR.
people may notice the metrics changed when upgrading to the new mosn.

jizhuozhi: I have modified to make it compatible. As for `UpstreamResponseFailed`, I think it should include all not ok (2xx) conditions, but it wasn't in the past, maybe that's worth discussing. cc @taoyuanyuan

doujiang24: nope, 3xx should not be consider as failed.

4xx could be consider as failed, but it's a break change, we could discuss it.

jizhuozhi: > nope, 3xx should not be consider as failed.

My mistake, of source 1xx/2xx/3xx never could be failed. Only 4xx/5xx or status code in configured set could be considered. But I cannot image which condition that 4xx/5xx occered but request is not failed.

doujiang24: I'm fine that 4xx to be considered failed, just a bit worried about the break change.

doujiang24: okay, we'd better keep backward compatibility.

jizhuozhi: Backward compatibility is no problem, and for those who need it, they can achieve the same effect with simple additions in the Dashboard.

But now failed has exactly the same meaning as server error, maybe we can mark `UpstreamResponse*` as deprecated and remove it in a future version

jizhuozhi: > okay, we'd better keep backward compatibility.

Sorry, my mistake, I realized failed not only occured in L7 but also L4. LGTM

pkg/proxy/downstream.go
```diff
@@ -1217,12 +1218,20 @@ func (s *downStream) onUpstreamHeaders(endStream bool) {
...
 func (s *downStream) handleUpstreamStatusCode() {
-	// todo: support config?
 	if s.upstreamRequest != nil && s.upstreamRequest.host != nil {
-		if s.requestInfo.ResponseCode() >= http.InternalServerError {
+		responseCode := s.requestInfo.ResponseCode()
+
+		switch {
+		case responseCode >= http.BadRequest && responseCode < http.InternalServerError:
+			s.upstreamRequest.host.HostStats().UpstreamResponseClientError.Inc(1)
+			s.upstreamRequest.host.ClusterInfo().Stats().UpstreamResponseClientError.Inc(1)
```
doujiang24, line 1227: missing `UpstreamResponseSuccess` here?

this looks a bit complicated now.

jizhuozhi: I will fix that

jizhuozhi: It's hard to decide, unless we give these keys a whole new meaning: `UpstreamResponseSuccess` and `UpstreamResponseFailed` represent L4 status, and HTTP status codes represent L7 status (like envoy https://www.envoyproxy.io/docs/ envoy/latest/configuration/http/http_conn_man/stats#statistics)

doujiang24: oh, I just mean the backward compatibility.

PR #2298, opened 2023-04-24
fix: `activeRequestBias` missing default value (1.0 but not 0.0), and fix typo

jizhuozhi:
The current LAR and LAC load balancing algorithms do not have a test for request distribution, adding tests to avoid untestable changes in the future

### Solutions

### UT result
Use a fixed weight and a fixed number of connections, but use a looser statistical error (delta)

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/upstream/cluster/lb_leastconnection_test.go
+++ b/pkg/upstream/cluster/lb_leastconnection_test.go
@@ -98,14 +98,14 @@ func TestLeastActiveConnectionLoadBalancer_ChooseHost(t *testing.T) {
 	t.Run("no bias", func(t *testing.T) {
 		verify(t, nil, 0.15)
 	})
-	t.Run("bias 0.9", func(t *testing.T) {
+	t.Run("low bias", func(t *testing.T) {
 		verify(t, &clusterInfo{
 			lbConfig: &v2.LbConfig{
 				ActiveRequestBias: 0.5,
 			},
 		}, 0.15)
 	})
-	t.Run("bias 1.1", func(t *testing.T) {
+	t.Run("high bias", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/upstream/cluster/lb_leastconnection_test.go
```diff
@@ -53,3 +56,60 @@ func TestLACChooseHost(t *testing.T) {
...
+		compareDistribution(t, expect, actual, delta)
+	}
+
+	t.Run("no bias", func(t *testing.T) {
+		verify(t, nil, 0.15)
+	})
+	t.Run("bias 0.9", func(t *testing.T) {
+		verify(t, &clusterInfo{
+			lbConfig: &v2.LbConfig{
+				ActiveRequestBias: 0.5,
```
spacewander, line 104: Why use 0.5 when the test name is "bias 0.9"?

jizhuozhi: my mistake, renamed

jizhuozhi:
Sorry, I found my mistake. When multiple branches are switched, some commits are not pushed. `activeRequestBias` has no default value, resulting in break change, the original semantic `activeRequestBias` is 1, but now it is 0.

I discovered the problem when these commits were rebased into this PR. I'm so sorry I almost caused a serious production accident.

cc @doujiang24 , @taoyuanyuan , @3062 , @spacewander

doujiang24 (dismissed):

taoyuanyuan:
pls resolve conflict.

taoyuanyuan (approved):

doujiang24 (approved):

PR #2286, opened 2023-04-06
feat: `LeastActiveRequest` (and `LeastActiveConnection`) supports setting the bias of active request metric

jizhuozhi:
### Issues associated with this PR

#2285 

### Solutions

Keep consistency between mosn and envoy

### UT result
Unit Test is needed if the code is changed, your unit test should cover boundary cases, corner cases, and some exceptional cases. And you need to show the UT result.

### Benchmark
If your code involves the processing of every request, you should give the Benchmark Result.

### Code Style
+ Make sure `Goimports` has run
+ Show `Golint` result

Issue #2285, opened 2023-04-06
`LeastActiveRequest` (and `LeastActiveConnection`) supports setting the bias of active request metric


jizhuozhi:
### The New Feature

`LeastActiveRequest` (and `LeastActiveConnection`) supports setting the bias of active request metric.

### Your scenes

In [envoy load balancers](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers), which supports configuring the bias of `activeRequest` to calculate actual weight

$$
Weight = \frac{LoadBalancingWeight} {(ActiveRequests + 1)^{ActiveRequestBias}}.
$$

### Your advice

Keep consistency between mosn and envoy, support convert from xds

### Environment

- MOSN Version

PR #2301, opened 2023-04-30
feat: PeakEWMA supports configuring activeRequestBias and optimizes the default EWMA of newly added hosts

jizhuozhi:

Since #2285 and #2286 makes `activeRequest` having a configurable bias, I want to make it effective in PeakEWMA

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


pkg/upstream/cluster/loadbalancer.go
```diff
@@ -779,10 +788,22 @@ func (lb *peakEwmaLoadBalancer) unweightedPeakEwmaScore(h types.Host) float64 {
...
+		duration = h.ClusterInfo().Stats().UpstreamRequestDurationEWMA.Rate()
+		// None of the active requests returned in cluster, or the metrics is disabled
+		if duration == 0 {
+			duration = float64(lb.defaultDuration)
+		}
+	}
+
+	biasedActiveRequest := float64(stats.UpstreamRequestActive.Count()) + 1
+	if biasedActiveRequest == 1.0 || lb.activeRequestBias == 0.0 {
+		biasedActiveRequest = 1
```
spacewander, line 803: Why set biasedActiveRequest to 1 when `if biasedActiveRequest == 1.0`?

jizhuozhi: avoid too many branches

spacewander: I take a look at Go's Pow implementation:
https://github.com/golang/go/blob/e335a2665f5e322a7da8baa22fe816b6ef9aaf24/src/math/pow.go#L57-L60
There are already additional special handling for these cases.

Do the additional branches in the caller's code improve the performance?

jizhuozhi: Thank you very much, I haven't paid attention to the implementation of Pow before, here is a redundant optimization

spacewander (approved):

doujiang24 (approved):

Issue #2303, opened 2023-05-05
EdfLoadBalancer supports configuring force weighted choose even hosts have same weights


jizhuozhi:
### The New Feature

EdfLoadBalancer provides a configure way to make sure user could force weighted choose even hosts have same weights.

### Your scenes

Since we have a dynamic weight function, it is not possible to directly determine whether all weights are the same by comparing weights, thus turning into unweighted choice.

For least request and PeakEWMA, hosts can be selected in constant time by P2C, but it is usually based on probability rather than weight (for A, B, C three nodes with different ordered weights, the probability of A is 2/3, the probability of B is 1/3, even if A is far better than B), but for real user scenarios, they will expect to choose as many as possible with good performance and avoid poor performance.

### Your advice

Provide a field `ForceWeighted` in `LbConfig` to make sure user could configure it.

### Environment

- MOSN Version

jizhuozhi:
It seems not a good idea that allow users configuring force weighted choose. 

When the probability of failure is 0.5, EDF is completely inferior to P2C, I am thinking whether I should use weighted P2C instead of EDF at all

<img width="827" alt="image" src="https://user-images.githubusercontent.com/31175030/236601178-006d1147-d26f-43bd-81f1-42822c62b41b.png">

PR #2304, opened 2023-05-05
feat: EdfLoadBalancer supports configuring force weighted choose even hosts have same weights

jizhuozhi:

#2303 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


jizhuozhi:
It seems not a good idea that using score as weight factor, this PR should be hold on
Image

jizhuozhi:
Since #2306 do not fix the weight of edf entry, it works correctly

image


When the probability of failure is 0.1, they behave the same, 

image

but when the probability of failure is 0.5, EDF is completely inferior to P2C, I am thinking whether I should use weighted P2C instead of EDF at all

image

PR #2306, opened 2023-05-06
bugfix: fix host weight in each `weightFunc` but not in `edfScheduler`

### LINKED ISSUE #2303: EdfLoadBalancer supports configuring force weighted choose even hosts have same weights

### LINKED ISSUE #2305: `edfscheduler` incorrectly fixes weights to 1 in dynamic load algorithm
### Describe the bug

Currently I try to allow users to force the use of EDF for weighted load balancing, even if all hosts have the same weight, because their actual weights are dynamic. But I found something interesting, they seem to have the same weight even though the performance distribution is completely different.

By logging the peeked edf entry, I found the weight are all be fixed 1 (https://github.com/mosn/mosn/issues/1969, https://github.com/mosn/mosn/issues/1970)

image

### Expected behavior

In my test suite, different servers have different latencies (following a normal distribution), where the best performer is assigned a weight of 2, and the other server has a failure injected with 0.1, which should be avoided as much as possible be chosen. In P2C (or Weighted P2C), it works correctly

image

### Actual behavior

the actual behavior is close to unweighted RoundRobin

image

### Minimal yet complete reproducer code (or GitHub URL to code)

Simulate back-end services with different response times, use least request or PeakEWMA, and assign them different weights, the final result is close to unweighted RoundRobin

### Environment

- MOSN Version since v0.27.0

jizhuozhi:

Fix #2303
Fix #2305

### Solutions
`weightFunc` is responsible for ensuring that the weight is legal, not adjusted by EDF, so we adjusted the logic of modifying hostWeight to each `weightFunc`, and `edfScheduler` is only responsible for scheduling

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -50,7 +50,6 @@ type WeightItem interface {
 
 // Add new item into the edfScheduler
 func (edf *edfScheduler) Add(item WeightItem, weight float64) {
-	weight = fixHostWeight(weight)
 	edf.lock.Lock()
 	defer edf.lock.Unlock()
 	entry := edfEntry{
--- a/pkg/upstream/cluster/edf.go
+++ b/pkg/upstream/cluster/edf.go
@@ -50,7 +50,7 @@ type WeightItem interface {
 
 // Add new item into the edfScheduler
 func (edf *edfScheduler) Add(item WeightItem, weight float64) {
-	weight = edfFixedWeight(weight)
+	weight = fixHostWeight(weight)
 	edf.lock.Lock()
 	defer edf.lock.Unlock()
 	entry := edfEntry{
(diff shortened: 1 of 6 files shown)


doujiang24 (dismissed):
better add an bad case as comment, for better understanding.

jizhuozhi:
In the dynamic weight load balancing algorithm, the result domain will be very wide. For example, in PeakEWMA, because Duration is used, it will be above 1e9 (nanoseconds). At this time, the weight divided by this value will be far less than 1 (at least 1e-8 ), when corrected at this time, it will become 1 by mistake, causing each host to have the same weight, which looks like RoundRobin

doujiang24 (approved):

spacewander (approved):

PR #2308, opened 2023-05-10
allow custom proxy use default route handler name

### LINKED ISSUE #2307: 需要移除proxy默认的route handler name
目前现状：
https://github.com/mosn/mosn/blob/80f7fa46133ba669137dc0184440463487eb3019/pkg/filter/network/proxy/factory.go#L120-L122
proxy是提供了默认的route handler名字: default
同时，route handler factory 提供了注册默认handler的方法：
https://github.com/mosn/mosn/blob/80f7fa46133ba669137dc0184440463487eb3019/pkg/router/factory.go#L33-L36

场景：
假设我自定义了个route handler, 名字为 test，同时设置为默认的route handler
那么proxy即使不声明route handler name，也会去找名为default的route handler，而不是我自定义的route handler

预期:
我注册了默认的route handler，理论上只要不声明，都应该使用我声明的router handler

fibbery:

fix #2307 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/router/types.go
+++ b/pkg/router/types.go
@@ -47,7 +47,11 @@ var (
 	ErrUnexpected           = errors.New("an unexpected error occurs")
 )
 
-var DefaultRouteHandlerName = types.DefaultRouteHandler
+var defaultRouteHandlerName = types.DefaultRouteHandler
+
+func GetDefaultRouteHandlerName() string {
+	return defaultRouteHandlerName
+}
 
 type headerFormatter interface {
 	format(ctx context.Context) string
@@ -162,7 +166,7 @@ func (f *handlerFactories) add(name string, h MakeHandlerFunc, isDefault bool) {
 	defer f.mutex.Unlock()
 	f.factories[name] = h
 	if isDefault {
-		DefaultRouteHandlerName = name
+		defaultRouteHandlerName = name
 		f.defaultFactory = h
 	}
 }
--- a/pkg/router/types.go
+++ b/pkg/router/types.go
@@ -47,6 +47,8 @@ var (
 	ErrUnexpected           = errors.New("an unexpected error occurs")
 )
 
+var DefaultRouteHandlerName = types.DefaultRouteHandler
+
 type headerFormatter interface {
 	format(ctx context.Context) string
 	append() bool
@@ -160,6 +162,7 @@ func (f *handlerFactories) add(name string, h MakeHandlerFunc, isDefault bool) {
 	defer f.mutex.Unlock()
 	f.factories[name] = h
 	if isDefault {
+		DefaultRouteHandlerName = name
 		f.defaultFactory = h
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/types/route.go
```diff
@@ -75,7 +75,7 @@ const (
 	HandlerStop
 )
 
-const DefaultRouteHandler = "default"
```
jizhuozhi, line 78: maybe it should not be exposed by variable, but by methods

doujiang24: I'm not sure introduce method in types package is a good idea.
or we could introduce a new global variable somewhere?

jizhuozhi: we could provide a new variable which has high priority, if it unset, using `DefaultRouteHandler` as default

pkg/router/types.go
```diff
@@ -47,6 +47,8 @@ var (
 	ErrUnexpected           = errors.New("an unexpected error occurs")
 )
 
+var DefaultRouteHandlerName = types.DefaultRouteHandler
```
jizhuozhi, line 50: There is no need to export this variable, please provide two methods

jizhuozhi: If static initialized, what would happen if it's unset to `""` ?

jizhuozhi (changes requested):

pkg/router/types.go
```diff
@@ -180,6 +184,13 @@ func (f *handlerFactories) exists(name string) bool {
 	return ok
 }
 
+func GetDefaultRouteHandler() string {
+	if GetMakeHandlerFunc(defaultRouteHandlerName) == nil {
```
jizhuozhi, line 188: maybe no need to get handler here, just check `defaultRouteHanlderName` is not empty string is enough

jizhuozhi: cc @nejisama

doujiang24: or, just set `defaultRouteHandlerName = types.DefaultRouteHandler` at line 51?

jizhuozhi: If static initialized, what would happen if it's unset to "" ? (As above comment)

jizhuozhi: Sorry, @fibbery , It's my mistake, there is no condition that user could reset handler name to empty string since it's not exported. 

So there is no need to use this method, static initializing `defaultRouteHandlerName = types.DefaultRouteHandler` is LGTM

doujiang24:
@fibbery perfer static initializing from my side, it could be simplier and simplier is better.

jizhuozhi:
> I think this way is also ok,  what do you think

Hello, @fibbery . As @doujiang24 said, perfer static initializing

doujiang24 (approved):

jizhuozhi (approved):
LGTM

jizhuozhi:
Hello, @fibbery . How about renaming this PR to `allow custom proxy default route handler name`?

jizhuozhi:
Maybe it's a question of my expression, this PR is LGTM, what I want to confirm is if there is a better name that reflects what this PR does? 

Because as we can see, after the PR changes, it doesn't remove anything, it just allows setting a variable with higher priority to override the default name when adding.

PR #2309, opened 2023-05-11
make grpc filter support uds

wenxuwan:

make grpc filter support unix

https://github.com/mosn/layotto/issues/867

### Solutions

根据network的类型，解析address，让grpc filter支持uds。

### 测试结果

server端配置：

image

client代码：

image

打印结果：

image

### Benchmark

### Code Style
+ Show `Golint` result


pkg/filter/network/grpc/factory.go
```diff
@@ -220,14 +221,14 @@ type Handler struct {
 }
 
 // New a grpc server with address. Same address returns same server, which can be start only once.
-func (s *Handler) New(addr string, conf json.RawMessage, options ...grpc.ServerOption) (*registerServerWrapper, error) {
```
jizhuozhi, line 223: It seems a break change, how about providing another method likes `NewUnix` (just an example name)?

pkg/filter/network/grpc/listener.go
```diff
@@ -34,13 +34,16 @@ type Listener struct {
 	addr    net.Addr
 }
 
-func NewListener(address string) (*Listener, error) {
```
jizhuozhi, line 37: As above

jizhuozhi (changes requested):

pkg/filter/network/grpc/factory.go
wenxuwan: I checked the code of Mosn. The changes of these two functions will not cause compatibility problems. Adding new functions will lead to code duplication and redundancy. I personally think that it may be more reasonable to add a parameter if there is no compatibility problem. 😊

jizhuozhi: But there has an special using case: users could use mosn as package, like layotto, without mosnd.

As for code duplication and redundancy, we could use a single method for them all, the exposed methods are just adaptor for your functions

nejisama: > I checked the code of Mosn. The changes of these two functions will not cause compatibility problems. Adding new functions will lead to code duplication and redundancy. I personally think that it may be more reasonable to add a parameter if there is no compatibility problem. 😊

Add an unexport function that have network.
 New() wraps tcp, NewUnix wraps unix.
No code duplication

jizhuozhi: Oooops，there is some lint failes

```
  string `unix` has 3 occurrences, make it a constant (goconst)
  
  Raw Output:
  filter/network/grpc/factory.go:88:16: string `unix` has 3 occurrences, make it a constant (goconst)
  	if network == "unix" {
  	              ^
  Error: [golangci] reported by reviewdog 🐶
  variable 'network' is only used in the if-statement (filter/network/grpc/factory.go:88:2); consider using short syntax (ifshort)
  
  Raw Output:
  filter/network/grpc/factory.go:76:2: variable 'network' is only used in the if-statement (filter/network/grpc/factory.go:88:2); consider using short syntax (ifshort)
  	network := cfg.Network
  	^
```

Unix and tcp could define a new type and mark them as const enums

wenxuwan: fixed

pkg/filter/network/grpc/var.go
```diff
@@ -22,6 +22,10 @@ import (
 	"mosn.io/pkg/variable"
 )
 
+const (
```
jizhuozhi, line 25: maybe it should be the top this file after package (reorder), and named as NetworkUnix and NetworkTcp (var name)

Otherwise LGTM

jizhuozhi: Sorry mymistake. Ignore that

wenxuwan: > Sorry mymistake. Ignore the first, please rename variables

1. I don't want to set the two variables as public here
2. Maybe unixNetwork has better semantic information than NetworkUnix?

jizhuozhi: It's ok that not exported, but those with the same intent preferably have the same prefix, like in C API they both have `AF_` prefix `syscall.AF_UNIX` and `syscall.AF_INET`

jizhuozhi (approved):
LGTM

doujiang24 (approved):

Issue #2317, opened 2023-05-26
Mosn failed to restart because the uds file already exists


wenxuwan:
### Describe the bug

The uds socket file will not be deleted when the Mosn exits abnormally, causing the Mosn to fail to restart

### Expected behavior

Mosn start success.

### Actual behavior

Mosn start fail.

### Steps to reproduce

1. configure file:

![image](https://github.com/mosn/mosn/assets/11661906/d5df15cd-bfdd-4b37-87b1-1ee5c936306d)

2. After Mosn started, sock file created in /tmp

![image](https://github.com/mosn/mosn/assets/11661906/89dd1034-827b-426c-8be4-7a5d57129f5c)

3. kill Mosn with -9

![image](https://github.com/mosn/mosn/assets/11661906/05df7c7a-ea5b-40ed-a7e3-ce8594bbab0d)

4. socket file still there

![image](https://github.com/mosn/mosn/assets/11661906/730f22ec-40c7-49bc-85f9-653e42a77d9e)

5. Mosn start fail

![image](https://github.com/mosn/mosn/assets/11661906/8ceb0733-1dc6-4a5f-976f-d05602988491)

### Environment

- MOSN Version
latest

PR #2318, opened 2023-05-26
remove uds socket file before unix listener start

wenxuwan:

#2317 

### Solutions

try to remove unix socket file before unix listener start


spacewander (commented):
Let's add a test in https://github.com/mosn/mosn/blob/master/pkg/network/listener_test.go

doujiang24 (approved):

spacewander (approved):

PR #2323, opened 2023-05-30
feat: fix duration when response error occurred

jizhuozhi:

#2295 

### Solutions
PeakEWMA records the response time to select the optimal host, but may select the wrong host when it fails fast. Record the EWMA of the total number of responses, client errors, and server errors, and use it to calculate the decaying failure rate, thereby correcting the response time for fast failures.

TODO: supports configuring bias for client errors and server errors.

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


doujiang24 (approved):

antJack (approved):
LGTM

PR #2348, opened 2023-10-09
[optimize] support set reuseport option to UDP listener and connection

dengqian:

When the UDP listener is bidirectional data transmission listener, multiple go routine will write the same socket to send packet to clients, which will cause  high softirqs.

### Solutions
So we change our implements to generate a new UDP socket and connect to client remote address, and use this socket to fulfill the subsequent data transmission.

This is the CPU usage comparision:

### DIFF
--- a/pkg/config/v2/server.go
+++ b/pkg/config/v2/server.go
@@ -63,6 +63,7 @@ type ListenerConfig struct {
 	Type                  ListenerType        `json:"type,omitempty"`
 	AddrConfig            string              `json:"address,omitempty"`
 	BindToPort            bool                `json:"bind_port,omitempty"`
+	ReusePort             bool                `json:"reuseport,omitempty"` // recommended to use in UDP bidirectional data transmission scenarios.
 	Network               string              `json:"network,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/config/v2/server.go
```diff
@@ -63,6 +63,8 @@ type ListenerConfig struct {
 	Type                  ListenerType        `json:"type,omitempty"`
 	AddrConfig            string              `json:"address,omitempty"`
 	BindToPort            bool                `json:"bind_port,omitempty"`
+	// recommended to use in UDP bidirectional data transmission scenarios.
+	ReusePort             bool                `json:"reuseport,omitempty"`
```
spacewander, line 67: Let's run gofmt in this file

spacewander (approved):

spacewander (approved):

spacewander:
@dengqian 

> This is the CPU usage comparision

The comparison graph does not display well, could you resubmit it?

dengqian:
> @dengqian
> 
> > This is the CPU usage comparision
> 
> The comparison graph does not display well, could you resubmit it?

done

doujiang24 (approved):

PR #2349, opened 2023-10-11
[opt] reduce UDP read buffer size

dengqian:

The default UDP read buffer size is now set to UDP packet max size, which is 64*1024 bytes, when the number of UDP connection increases, the memory usage will increase rapidly

### Solutions
Enable using the "default_read_buffer_size" configuration value as read buffer size and change the default UDP read buffer size to 4096

### DIFF
--- a/pkg/network/connection.go
+++ b/pkg/network/connection.go
@@ -46,7 +46,7 @@ import (
 
 // Network related const
 const (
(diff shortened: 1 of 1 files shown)


3062 (approved):

taoyuanyuan (approved):
LGTM

Issue #2368, opened 2023-12-20
recycle a bad connect to http1 connPool


alpha-baby:
### Describe the bug

![image](https://github.com/mosn/mosn/assets/47237632/b91dbe6e-1a59-4191-a700-3bb83ac6f252)
![image](https://github.com/mosn/mosn/assets/47237632/04974b82-b11a-4adf-8664-f2e29345b090)

### Steps to reproduce

first：

start sample http server

```golang
package main

import (
	"fmt"
	"math/rand"
	"net/http"
	"time"
)

// 根据长度生成一个随机字符串
func RandString(length int) string {
	str := "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
	bytes := []byte(str)
	result := []byte{}
	r := rand.New(rand.NewSource(time.Now().UnixNano()))
	for i := 0; i < length; i++ {
		result = append(result, bytes[r.Intn(len(bytes))])
	}
	return string(result)
}

func ServeHTTP(w http.ResponseWriter, r *http.Request) {
	fmt.Printf("[UPSTREAM]receive request %s", r.URL)
	fmt.Println()

	w.Header().Set("Content-Type", "text/plain")
	w.Header().Set("MaxLengthHeaderKey", RandString(100))

	fmt.Fprintf(w, "Method: %s\n", r.Method)
	fmt.Fprintf(w, "Protocol: %s\n", r.Proto)
	fmt.Fprintf(w, "Host: %s\n", r.Host)
	fmt.Fprintf(w, "RemoteAddr: %s\n", r.RemoteAddr)
	fmt.Fprintf(w, "RequestURI: %q\n", r.RequestURI)
	fmt.Fprintf(w, "URL: %#v\n", r.URL)
	fmt.Fprintf(w, "Body.ContentLength: %d (-1 means unknown)\n", r.ContentLength)
	fmt.Fprintf(w, "Close: %v (relevant for HTTP/1 only)\n", r.Close)
	fmt.Fprintf(w, "TLS: %#v\n", r.TLS)
	fmt.Fprintf(w, "\nHeaders:\n")

	r.Header.Write(w)
}

func main() {
	http.HandleFunc("/", ServeHTTP)
	http.ListenAndServe("127.0.0.1:8080", nil)
}

```

second：

start mosn with json config as follows

```json
{
	"disable_upgrade" : true,
	"servers":[
		{
			"default_log_path":"stdout",
[...]

PR #2369, opened 2023-12-22
fix: set max_header_size not effect, not recycle a bad conn for http1

alpha-baby:

#2328 #2368 

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result

### DIFF
--- a/pkg/stream/http/connpool.go
+++ b/pkg/stream/http/connpool.go
@@ -345,6 +345,12 @@ func (ac *activeClient) OnResetStream(reason types.StreamResetReason) {
 				ac.client.ConnID())
 		}
 		ac.closeConn = true
+	} else if reason == types.StreamRemoteReset && !ac.closed {
+		if log.DefaultLogger.GetLogLevel() >= log.DEBUG {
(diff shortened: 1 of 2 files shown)


spacewander (dismissed):

pkg/stream/http/connpool.go
```diff
@@ -345,6 +345,12 @@ func (ac *activeClient) OnResetStream(reason types.StreamResetReason) {
 				ac.client.ConnID())
 		}
 		ac.closeConn = true
+	} else if reason == types.StreamRemoteReset && !ac.closed {
```
wangfakang, line 348: Good catch.  But whether the value of variable `reason` can also be set correctly in the exception. 
https://github.com/mosn/mosn/blob/625bfad6df6f760d3e4b48bd897ef7f7db749cdf/pkg/stream/http/stream.go#L292-L300

alpha-baby:
先不要合并，此 PR 会作为一个示例给别人分享怎么参与开源项目

doujiang24 (approved):
lgtm

spacewander (approved):

Issue #2374, opened 2024-01-05
The ClientErrorBias and ServerErrorBias in lbconfig are not using the default values correctly


3062:
### Describe the bug

#2337 
If LbConfig is configured but ClientErrorBias and ServerErrorBias are not, then 0 will be used instead of the default value

jizhuozhi:
The current load balancing configuration (lbconfig) is really confusing, and the logic of various default values ​​is unclear. I think this area really needs to be modified in a unified way.

But the difficulty here itself is not high, I think it is very suitable as a good first issue

NX-Official:
I will fix this issue in this days❤️

PR #2375, opened 2024-01-07
fix: lbconfig use default values correctly

NX-Official:

#2374 

### Solutions

use `*float` instead of `float`, so when it is `nil` after `json.Unmarshal()`, we can manually load the default value


3062 (approved):

pkg/upstream/cluster/loadbalancer.go
```diff
@@ -816,3 +818,19 @@ func (lb *peakEwmaLoadBalancer) unweightedPeakEwmaScore(h types.Host) float64 {
 
 	return duration * biasedActiveRequest / successRate
 }
+
+// loadConfigValueFloat64 If it is nil after JSON.Unmarshal, it will be set to default value
+func loadConfigValueFloat64(configOption *float64, defaultValue float64) float64 {
```
jizhuozhi, line 823: Maybe this function could move to `util` or other common package

jizhuozhi (approved):
LGTM

3062 (approved):

jizhuozhi (approved):
LGTM

PR #2393, opened 2024-05-26
fix: fix miss makezero bug

alingse:
### Issues associated with this PR

Your PR should present related issues you want to solve.

I was running github actions to run linter [makezero](https://github.com/ashanbrown/makezero) for top github golang repos.

see issues https://github.com/alingse/go-linter-runner/issues/1

and the github actions output  https://github.com/alingse/go-linter-runner/actions/runs/9242656489/job/25425751948

```
====================================================================================================
append to slice `frag` with non-zero initialized length at https://github.com/mosn/mosn/blob/master/pkg/module/http2/mhttp2.go#L1[9](https://github.com/alingse/go-linter-runner/actions/runs/9242656489/job/25425751948#step:4:10)38:10
====================================================================================================
```

### Solutions
You should show your solutions about the issues in your PR, including the overall solutions, details and the changes. At this time, Chinese is allowed to describe these.

### UT result
Unit Test is needed if the code is changed, your unit test should cover boundary cases, corner cases, and some exceptional cases. And you need to show the UT result.

### Benchmark
If your code involves the processing of every request, you should give the Benchmark Result.

### Code Style
+ Make sure `Goimports` has run
+ Show `Golint` result

PR #2411, opened 2024-10-02
fix: fix slice init length

cuishuang:

The intention here should be to initialize a slice with a capacity of 1 rather than initializing the length of this slice.

### Solutions

### UT result

### Benchmark

### Code Style
+ Show `Golint` result


taoyuanyuan (approved):
LGTM

taoyuanyuan:
same as https://github.com/mosn/mosn/pull/2393
thank you~

